PR #35, opened 2022-02-23
misc(config/zkSyncDeploy/ethNetwork): shows where RPC can be provided

JayWelsh:
Right now, if anyone uses the default setup shown in the tutorials within the documentation, they will be using a heavily throttled default API key:

Screenshot 2022-02-23 at 15 21 51

I don't believe it is clear or obvious how to avoid this, so I think adding these comments from this PR into the example code will make it much easier for developers to avoid running into this throttling (by pointing out where their own RPC address can be provided).

### DIFF
--- a/docs/api/hardhat/compiling-libraries.md
+++ b/docs/api/hardhat/compiling-libraries.md
@@ -80,7 +80,7 @@ module.exports = {
   },
   zkSyncDeploy: {
     zkSyncNetwork: "https://zksync2-testnet.zksync.dev",
-    ethNetwork: "goerli", // Can also be RPC address of network (e.g. `https://goerli.infura.io/v3/<API_KEY>`)
+    ethNetwork: "goerli", // Can also be the RPC URL of the network (e.g. `https://goerli.infura.io/v3/<API_KEY>`)
   },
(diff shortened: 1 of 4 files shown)


StanislavBreadless (approved):

docs/api/hardhat/compiling-libraries.md
```diff
@@ -80,7 +80,7 @@ module.exports = {
   },
   zkSyncDeploy: {
     zkSyncNetwork: "https://zksync2-testnet.zksync.dev",
-    ethNetwork: "goerli",
+    ethNetwork: "goerli", // Can also be RPC address of network (e.g. `https://goerli.infura.io/v3/<API_KEY>`)
```
StanislavBreadless, line 83: Maybe this way is better?

```suggestion
    ethNetwork: "goerli", // Can also be the RPC URL of the network (e.g. `https://goerli.infura.io/v3/<API_KEY>`)
```

JayWelsh:
Hi @StanislavBreadless , I've made the suggested changes :)

StanislavBreadless (approved):

PR #52, opened 2022-03-21
Update hardhat docs

ly0va:


docs/api/hardhat/reference.md
```diff
@@ -24,30 +24,32 @@ This plugin most often will not be used directly in the code.
 
 ### Configuration
 
-```js
+```json
```
StanislavBreadless, line 27: why?

The code below is not a valid json.

ly0va: Also not a valid javascript though. I can make it a valid one if you have a strong preference, but it looks ok either way.

StanislavBreadless, line 27:
```suggestion
```js
```

StanislavBreadless (approved):

PR #65, opened 2022-06-22
Fixes to new release

vladbochok:


docs/api/api.md
```diff
@@ -175,7 +175,7 @@ Otherwise, the object of the following format is returned:
 }
 ```
 
-The `id` is the position of the leaf in the Merkle tree of L2->L1 messages for the block. The `proof` is the merkle proof for the message, while the `root ` is the root of the Merkle tree of L2->L1 messages. Please note, that the merkle tree uses *sha256* for the trees.
+The `id` is the position of the leaf in the Merkle tree of L2->L1 messages for the block. The `proof` is the merkle proof for the message, while the `root ` is the root of the Merkle tree of L2->L1 messages. Please note, that the merkle tree uses _sha256_ for the trees.
```
StanislavBreadless, line 178: Just out of interest: why? :)

the ** and __ have the same effect

StanislavBreadless (approved):

docs/api/api.md
vladbochok: It is not me, I just run `yarn ci:fix` :)

PR #68, opened 2022-06-23
Tutorial for building an AA factory

StanislavBreadless:


docs/dev/zksync-v2/system-contracts.md
```diff
@@ -29,3 +29,9 @@ This contract is used to send messages from zkSync to Ethereum L1. For each mess
 [Interface](https://github.com/matter-labs/v2-testnet-contracts/blob/6a93ff85d33dfff0008624eb9777d5a07a26c55d/l2/system-contracts/interfaces/INonceHolder.sol#L5).
 
 This contract stores nonces for the accounts. The nonces for the accounts are stored in a centralized place for efficiency ([the tx nonce and the deployment nonce](./contracts.md#differences-in-create-behaviour) are stored in a single place) and also for the ease of the operator.
+
+## Bootloader
+
+For the greater extensibility and to lower the overhead, some parts of the protocol (e.g. account abstraction rules) were moved to an ephemeral contract called _bootloader_. We call it ephemeral since it is formally never deployed and it can not be called, but it has a formal [address](https://github.com/matter-labs/v2-testnet-contracts/blob/6a93ff85d33dfff0008624eb9777d5a07a26c55d/l2/system-contracts/Constants.sol#L19) that is used as `msg.sender` whenever it calls other contracts.
```
barakshani, line 35:
```suggestion
For greater extensibility and to lower the overhead, some parts of the protocol (e.g. account abstraction rules) were moved to an ephemeral contract called _bootloader_. We call it ephemeral since formally it is never deployed and can not be called, but it has a formal [address](https://github.com/matter-labs/v2-testnet-contracts/blob/6a93ff85d33dfff0008624eb9777d5a07a26c55d/l2/system-contracts/Constants.sol#L19) that is used on `msg.sender`, when it calls other contracts.

Comment (barakshani) [reply docs/dev/zksync-v2/system-contracts.md:35]: Are you sure "ephemeral" is the right word? Google says "lasting for a very short time.".
I suggest calling it a "formal contract"

File: docs/dev/zksync-v2/system-contracts.md [hunk 2]
```diff
@@ -29,3 +29,9 @@ This contract is used to send messages from zkSync to Ethereum L1. For each mess
...
 
 This contract stores nonces for the accounts. The nonces for the accounts are stored in a centralized place for efficiency ([the tx nonce and the deployment nonce](./contracts.md#differences-in-create-behaviour) are stored in a single place) and also for the ease of the operator.
+
+## Bootloader
+
+For the greater extensibility and to lower the overhead, some parts of the protocol (e.g. account abstraction rules) were moved to an ephemeral contract called _bootloader_. We call it ephemeral since it is formally never deployed and it can not be called, but it has a formal [address](https://github.com/matter-labs/v2-testnet-contracts/blob/6a93ff85d33dfff0008624eb9777d5a07a26c55d/l2/system-contracts/Constants.sol#L19) that is used as `msg.sender` whenever it calls other contracts. 
+
+For now, you do not need to know any details about it, but knowing that it exists is important when you develop using the account abstraction feature. You can always assume that the bootloader is not malicious and it is a part of the protocol. In the future, the code of the bootloader will be made public and any changes to it will also mean an upgrade of the protocol.
```
barakshani, line 37:
```suggestion
For now, you do not have to know any details about it, but knowing that it exists is important when you develop using the account abstraction feature. You can always assume that the bootloader is not malicious and it is a part of the protocol. In the future, the code of the bootloader will be made public and any changes to it will also mean an upgrade to the protocol.

Comment (barakshani) [reply docs/dev/zksync-v2/system-contracts.md:37]: > You can always assume that the bootloader is not malicious

mmmm..... let's rephrase this

File: docs/dev/guide/custom-aa-tutorial.md [hunk 3]
```diff
@@ -0,0 +1,637 @@
...
+
+To increment the nonce, you should use the `incrementNonceIfEquals` method of the `NONCE_HOLDER_SYSTEM_CONTRACT` system contract. It accepts the nonce of the transaction and checks whether the nonce is the same as the provided one. If not, the transaction reverts. Otherwise, the nonce is increased.
+
+Even though the requirements above allow the accounts to touch only their storage slots, accessing your nonce in the `NONCE_HOLDER_SYSTEM_CONTRACT` is a whitelisted case, since it behaves in the same way as your storage, it just happened to be in another contract. To call the `NONCE_HOLDER_SYSTEM_CONTRACT`, you should add the folloing import:
+
+```solidity
+import '@matterlabs/zksync-contracts/l2/system-contracts/Constants.sol';
+```
+
+The `TransactionHelper` library (already imported in the example above) can be used to get the hash of the transaction that should be signed. You can also implement your own signature scheme and use different a commitment for the transaction to sign, but in this example, we will use the hashes provided by this library.
```
barakshani, line 147:
```suggestion
The `TransactionHelper` library (already imported in the example above) can be used to get the hash of the transaction that should be signed. You can also implement your own signature scheme and use a different commitment for the transaction to sign, but in this example we use the hash provided by this library.

File: docs/dev/guide/custom-aa-tutorial.md [hunk 4]
```diff
@@ -0,0 +1,637 @@
...
+    }
+
+    // Needed for the transaction to be correctly processed by the server.
+    require(success);
+}
+```
+
+Note, that whether the operator will consider the transaction successful will depend only on whether or not the call to `executeTransactions` was successful. So, it is highly recommended to put `require(success)` for the transaction, so that the users could get the best UX.
+
+### Full code for the AA
```
barakshani, line 191:
```suggestion
### Full code for AA

File: docs/dev/guide/custom-aa-tutorial.md [hunk 5]
```diff
@@ -0,0 +1,637 @@
...
+        address owner2
+    ) external returns (address) {
+        return DEPLOYER_SYSTEM_CONTRACT.create2AA(salt, aaBytecodeHash, 0, abi.encode(owner1, owner2));
+    }
+}
+```
+
+Note, that on zkSync, the deployment is not done via bytecode, but via bytecode hash. The bytecode itself is passed to the operator via `factoryDeps` field. Note, that the `_aaBytecodeHash` must be formed in a special way:
+
+- Firstly, it is hashed with sha256
```
barakshani, line 315:
```suggestion
- Firstly, it is hashed with sha256.

File: docs/dev/guide/custom-aa-tutorial.md [hunk 6]
```diff
@@ -0,0 +1,637 @@
...
+- Then, the first two bytes are replaced with the length of the bytecode in 32-byte words.
+
+You don't need to worry about it, since our SDK provides a built-in method to do it, that will be explained below.
```
barakshani, line 318:
```suggestion
You don't need to worry about it, since our SDK provides a built-in method to do it, explained below.

File: docs/dev/guide/custom-aa-tutorial.md [hunk 7]
```diff
@@ -0,0 +1,637 @@
...
+
+To deploy a factory, we need to create a deployment script. Create the `deploy` folder and create one file there: `deploy-factory.ts`. Put the following deployment script there:
+
+```ts
+import { utils, Wallet } from "zksync-web3";
+import * as ethers from "ethers";
+import { HardhatRuntimeEnvironment } from "hardhat/types";
+import { Deployer } from "@matterlabs/hardhat-zksync-deploy";
+
+// An example of a deploy script that will deploy and call a simple contract.
```
barakshani, line 330:
```suggestion
// An example of a deploy script that deploys and call a simple contract

File: docs/dev/guide/custom-aa-tutorial.md [hunk 8]
```diff
@@ -0,0 +1,637 @@
...
+export default async function (hre: HardhatRuntimeEnvironment) {
+    const wallet = new Wallet("<PRIVATE-KEY>");
+    const deployer = new Deployer(hre, wallet);
+    const factoryArtifact = await deployer.loadArtifact("AAFactory");
+    const aaArtifact = await deployer.loadArtifact("TwoUserMultisig");
+
+    // Deposit some funds to L2 in order to be able to perform L2 transactions.
+    // You can remove the depositing step if you know that the `wallet` has enough
+    // funds on zkSync.
```
barakshani, line 339:
```suggestion
    // You can remove the depositing step if the `wallet` has enough funds on zkSync

File: docs/dev/guide/custom-aa-tutorial.md [hunk 9]
```diff
@@ -0,0 +1,637 @@
...
+    const depositAmount = ethers.utils.parseEther("0.001");
+    const depositHandle = await deployer.zkWallet.deposit({
+        to: deployer.zkWallet.address,
+        token: utils.ETH_ADDRESS,
+        amount: depositAmount,
+    });
+    await depositHandle.wait();
+
+    // Getting the bytecodeHash of the AA
```
barakshani, line 348:
```suggestion
    // Getting the bytecodeHash of the account

File: docs/dev/guide/custom-aa-tutorial.md [hunk 10]
```diff
@@ -0,0 +1,637 @@
...
+        // Since the factory requires the code for the AA to be available,
+        // we should pass it here as well
+        aaArtifact.bytecode
+    ]);
+
+    console.log(`AA factory address: ${factory.address}`);
+}
+```
+
+In order to deploy the factory, you should run compile the contracts and run the script:
```
barakshani, line 361:
```suggestion
In order to deploy the factory, you should compile the contracts and run the script:

File: docs/dev/guide/custom-aa-tutorial.md [hunk 11]
```diff
@@ -0,0 +1,637 @@
...
+yarn hardhat deploy-zksync --script deploy-factory.ts
+```
+
+The output should be roughly the following:
+
+```
+AA factory address: 0x9db333Cb68Fb6D317E3E415269a5b9bE7c72627Ds
+```
+
+Where the address will be different for each run.
```
barakshani, line 374:
```suggestion
Note that the address will be different for each run.

File: docs/dev/guide/custom-aa-tutorial.md [hunk 12]
```diff
@@ -0,0 +1,637 @@
...
+
+## Working with accounts
+
+### Deploying account
```
barakshani, line 378:
```suggestion
### Deploying an account

File: docs/dev/guide/custom-aa-tutorial.md [hunk 13]
```diff
@@ -0,0 +1,637 @@
...
+
+Now, let's deploy an AA and initiate a new transaction with it. In this section we already assume that you have an EOA wallet with enough funds on it.
```
barakshani, line 380:
```suggestion
Now, let's deploy an account and initiate a new transaction with it. In this section we assume that you already have an EOA account with enough funds on it.

File: docs/dev/guide/custom-aa-tutorial.md [hunk 14]
```diff
@@ -0,0 +1,637 @@
...
+    );
+    console.log(`Deployed on address ${multisigAddress}`);
+}
+```
+
+_Note, that zkSync has different address derivation rules from Ethereum_. You should always use the `createAddress` and `create2Address` utility methods of the `zksync-web3` SDK.
+
+### Starting a transaction from this account
+
+Before the AA can do any transactions, we firsly need to top it up:
```
barakshani, line 437:
```suggestion
Before the deployes account can do any transactions, we need to top it up:

File: docs/dev/guide/custom-aa-tutorial.md [hunk 15]
```diff
@@ -0,0 +1,637 @@
...
+
+```ts
+await (await wallet.sendTransaction({
+    to: multisigAddress,
+    value: ethers.utils.parseEther('0.0001')
+})).wait();
+```
+
+Now, let's as an example try to deploy a new multisig, but the initiator of the transaction will be our account deployed in the previous part:
```
barakshani, line 446:
```suggestion
Now, as an example, let's try to deploy a new multisig, but the initiator of the transaction will be our deployed account in the previous part:

File: docs/dev/zksync-v2/system-contracts.md (hunk 2 above)
Comment (StanislavBreadless) [reply docs/dev/zksync-v2/system-contracts.md:37]: "is honest"?

File: docs/dev/guide/custom-aa-tutorial.md [hunk 16]
```diff
@@ -0,0 +1,637 @@
...
+    function validateTransaction(Transaction calldata _transaction) external payable override onlyBootloader {
+        _validateTransaction(_transaction);
+    }
+
+    function _validateTransaction(Transaction calldata _transaction) internal {
+        
+    }
+
+    function executeTransaction(Transaction calldata _transaction) external payable override onlyBootloader {
+		_executeTransaction(_transaction)
```
arose0, line 62: Missing `;`

docs/dev/guide/custom-aa-tutorial.md
```diff
@@ -0,0 +1,637 @@
...
+        
+    }
+
+    function executeTransactionFromOutside(Transaction calldata _transaction) external payable {
+		_validateTransaction(_transaction);
+		_executeTransaction(_transaction);
+	}
+
+
+	// bytes4(keccak256("isValidSignature(bytes32,bytes)")
```
arose0, line 75: Did you mean to leave this here but commented out?

docs/dev/guide/custom-aa-tutorial.md
```diff
@@ -0,0 +1,637 @@
...
+}
+```
+
+Note, that only the [bootloader](../zksync-v2/system-contracts.md#bootloader) should be allowed to call the `validateTransaction`/`executeTransaction` methods. That's why the `onlyBootloader` is used for them.
+
+The `executeTransactionFromOutside` is needed to allow external users to initiate transactions from this account. The easiest way to implement it is to do the same as `validateTransaction` + `executeTransaction` would do.
+
+### Signature validation
+
+Firstly, we need to implement the signature validation process. Since we are building a two-account multisig, let's pass its owners' addresses in the constructor. For signature validation, we will use Openzeppelin's utility libraries for dealing with `ECDSA`.
```
arose0, line 97: nit, Openzeppelin -> OpenZeppelin

docs/dev/guide/custom-aa-tutorial.md
StanislavBreadless: Yes, it just shows how the constant was derived. Maube we should say "EIP1271 magic constant"

StanislavBreadless: I'll probably just remove it

docs/dev/guide/custom-aa-tutorial.md
```diff
@@ -0,0 +1,637 @@
...
```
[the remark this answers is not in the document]
StanislavBreadless: Hmm, not sure, it is a tutorial. And only for this tutorial, will I use the ECDSA library. Generally, users could do whatever they want

docs/dev/guide/custom-aa-tutorial.md
barakshani: I think it's good

docs/dev/guide/custom-aa-tutorial.md
StanislavBreadless: I think it is better to put the `Full code for AA` here.

docs/dev/guide/custom-aa-tutorial.md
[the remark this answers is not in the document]
barakshani: My comment was that, in my opinion, it is better using present tense than future tense.

For your comment, I will then add: "In this tutorial we use...."

docs/dev/guide/custom-aa-tutorial.md
StanislavBreadless: I would say `deploys and calls`. Why remove the `.` at the end of the comment? Most of our comments end with a period

docs/dev/guide/custom-aa-tutorial.md
```diff
@@ -0,0 +1,637 @@
...
+        token: utils.ETH_ADDRESS,
+        amount: depositAmount,
+    });
+    await depositHandle.wait();
+
+    // Getting the bytecodeHash of the AA
+    const bytecodeHash = utils.hashBytecode(aaArtifact.bytecode);
+
+    const factory = await deployer.deploy(factoryArtifact, [bytecodeHash], undefined, [
+        // Since the factory requires the code for the AA to be available,
```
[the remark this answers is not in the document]
StanislavBreadless: I think "the" is more appropriate here, since we are talking a particular code

docs/dev/guide/custom-aa-tutorial.md
StanislavBreadless: On L2. From the user's perspective it is indistinguishable from an EOA, so I would avoid saying "EOA-like" to avoid confusion

docs/dev/guide/custom-aa-tutorial.md
```diff
@@ -0,0 +1,637 @@
...
+        ergsPerPubdata: '1',
+        feeToken: utils.ETH_ADDRESS
+    } as Eip712Meta,
+    value: ethers.BigNumber.from(0)
+}
+```
+
+::: tip Note on gasLimit
+
+Currently, we expect the `gasLimit` to cover both the verification and the execution step. Currently, the number of ergs that is returned by the `estimateGas` is `execution_ergs + 20000`, where `20000` is roughly equals to the overhead needed for the defaultAA to have both fee charged and the signature verified. In case your AA has very expensive verification step, you should some constant to the `gasLimit`.
```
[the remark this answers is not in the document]
StanislavBreadless: yeah :) "add"

docs/dev/guide/custom-aa-tutorial.md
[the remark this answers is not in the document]
barakshani: Yes, I am not suggestion to remove the "the" before "code"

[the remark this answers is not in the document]
StanislavBreadless: but you suggest to remove it from "AA". I meant that the code for a particular AA

docs/dev/guide/custom-aa-tutorial.md
barakshani: well, I generally agree, and for me this is understandable. However, I do think that we generate more confusion here, as on the one hand we talk about abstractions of accounts -- there are no more EOA _or_ contracts -- but on the other we keep using this terminology.

Maybe "default EOA logic"? or "default account logic"

StanislavBreadless: No, I would say that we have three citizens here:

- EOA
- Contract
- AA

The EOA *might* get different behavior, since we want to support `tx.origin` and if we want to preserve the semantics of `tx.origin == msg.sender`, then we'll need to keep it different

docs/dev/guide/custom-aa-tutorial.md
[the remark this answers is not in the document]
barakshani: Generally, people read abbreviations as they are so AA is "/ˈeɪ/ /ˈeɪ/", and we don't put "the" before that.

Personally I read AA as "account abstraction", and also in that case we don't put "the" (we don't say "the abstraction").

docs/dev/guide/custom-aa-tutorial.md
StanislavBreadless: Words like "default EOA logic" or "default account logic" imply that the user had to deploy such a wallet, but they don't need to

barakshani: Doesn't the first 2 just specific instationations of AA?

barakshani: mmm... default means default -- nothing particular to do. I think

docs/dev/guide/custom-aa-tutorial.md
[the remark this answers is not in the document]
StanislavBreadless: let's then say "the account"?

docs/dev/guide/custom-aa-tutorial.md
StanislavBreadless: > Doesn't the first 2 just specific instationations of AA?

We can put it this way :) 

> mmm... default means default -- nothing particular to do. I think

I just don't like that describing it takes 3 words, sounds too complicated for me

barakshani: your call

barakshani (approved):

PR #80, opened 2022-07-23
Finality paragraph update

kuzivany:
Removed the word `that` so the sentence makes more sense:
> the more L1 blocks generated after the initial block is processed&hellip;

Alternatively could've had:
> the more L1 blocks _that are_ generated after the initial block is processed&hellip;

Prefer the former as there are less words & possibly less cognitive load


githubdoramon:
docs were updated, so closing this as deprecated

PR #87, opened 2022-08-26
add steps to verify contracts

krofax:


docs/dev/guide/contract-verification.md
```diff
@@ -0,0 +1,86 @@
...
+The bytecode generated by the Solidity compiler is essentially deployed when a Smart Contract is deployed on a network. The original Solidity code is not retained in any way.
+
+Verifying a contract entails determining whether the on-chain bytecode generated from the Smart Contract source code complies. If it does, we can state that we have confirmed the bytecode's source code's integrity. Hence the name – verification.
+<br>
+
+<hr>
+
+<br>
+
+## How are Smart Contracts verified in zkSync??
```
barakshani, line 14: why 2 question marks?

docs/dev/guide/contract-verification.md
```diff
@@ -0,0 +1,86 @@
...
+<hr>
+
+<br>
+
+## Modes of verification
+
+Depending on whether your contract is deployed on a network or with zkSync, there are two modes of contract verification:
+
+- Public: Every zkSync user on the block explorer will be able to check the verified source code and interact with it. 
+- Private: The contract is verified and thus valid within a single project in zkSync, available to you and anybody with access to the project.
```
barakshani, line 41: I don't understand this part

docs/dev/guide/contract-verification.md
```diff
@@ -0,0 +1,86 @@
...
+<br>
+
+### Enter contract details
+
+To verify the contract, enter the following details;
+
+- Contract address: The address provided must match with `0x` address supplied during contract creation.
+- Contract name: The name must be identical to the name provided in the contract.
+- Optimization: This checks whether or not optimization was used while compiling the contract, If you enabled optimization during compilation, check **Yes** otherwise select **No**
+- Compiler version: This specifies the exact version of the compiler used to compile a Smart Contract. Click on the dropdown to specify the compiler version used.
```
barakshani, line 65: we need to explain somewhere how to obtain the compiler version

docs/dev/guide/contract-verification.md
krofax: This section just states that they are generally only two modes of smart contract verification, depending on what network it is verified, however, if the user verifies a smart contract using zkSync block explorer, it's made public to everyone to see and interact with.

docs/dev/guide/contract-verification.md
krofax: Screenshot 2022-08-26 at 13 34 44

I'm not sure how to go about this.

From your comment, you said "the explorer would offer to choose the compiler version", how then can users programmatically obtain the compiler version asides from what the explorer provides?

docs/dev/guide/contract-verification.md
krofax: A typo on my end, I'll correct this.

docs/dev/guide/contract-verification.md
barakshani: These tasks are completed by now, so there's no need to go in depth about it (what you have there already seems fine to me).

My comment was that we need to specify to devs to be aware of the compiler version they use while compiling their contracts, so they would know which version to use on the block explorer.

Once this is completed, we should add some text about how to retrieve it from artifacts:
https://linear.app/matterlabs/issue/SMA-1199/add-compiler-version-to-artifacts

docs/dev/guide/contract-verification.md
barakshani: what is "deployed on a network or with zkSync"?

PR #92, opened 2022-09-06
System contracts: update commit hashes

popzxc:
We've had outdated ones.

### DIFF
--- a/docs/dev/zksync-v2/system-contracts.md
+++ b/docs/dev/zksync-v2/system-contracts.md
@@ -6,13 +6,13 @@ Until the code for the system contracts has gone thorough testing, the code for 
 
 ## Interfaces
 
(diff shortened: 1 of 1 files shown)


StanislavBreadless:
@popzxc maybe it is better to do it like [here](https://github.com/matter-labs/zksync-web-v2-docs/pull/93), i.e. do not use commits, but relative path as commits get outdated with each upgrade so that we have less links to maintain

popzxc:
> but relative path as commits get outdated with each upgrade so that we have less links to maintain

It works both ways: if we push something to that repo before the testnet upgdade (i.e. when/if we'll have multiple envs), these links may also become incorrect. Think, for example, mainnet having one version of these contracts and testnet having another. I think hard revisions are better, though yeah they require more support.

StanislavBreadless:
@popzxc okay, up to you

StanislavBreadless (approved):

PR #96, opened 2022-09-12
Describe L2 blocks and L1 batches

perekopskiy:


docs/dev/zksync-v2/blocks-and-time.md
```diff
@@ -1,3 +1,21 @@
 # Block numbers and time
 
-TODO.
+On ZkSync there are two types of blocks: L2 blocks and L1 batches.
+
+## L2 blocks
+
+L2 block is similar to a common block on L1.
+L2 block time is almost constant, like a few seconds.
+In this documentation by default the word "block" actually
```
popzxc, line 9: Lines seem to be unnecessarily short.

docs/dev/zksync-v2/blocks-and-time.md
```diff
@@ -1,3 +1,21 @@
...
+means "L2 block" including API.
+L2 block hash is deterministic
+and is basically equal to `keccak256(bytes4(l2_block_number))`.
+
+## L1 batches
+
+L1 batch is a range of L2 blocks that are committed
+and finalized on L1 as one whole.
+L1 batch time can be different: from minutes to several hours
```
popzxc, line 18: "depending on the system load" is an obscure statement. I think you should clarify that to achieve cost optimization we try use the whole capacity we have, and generally we only try to seal the batch when it's full, whereas "cheap" l2 blocks can be sealed quickly.

popzxc (commented):
This description seems to be correct, but not very helpful or interesting for users. I guess the general advice would be: try to imagine that you're user who knows nothing about the topic, and you really want to know what is going on and *why*. Try to imagine questions you'd have while reading the doc, and try to answer them in a way that is accessible and (ideally) interesting to follow.

docs/dev/zksync-v2/blocks-and-time.md
```diff
@@ -1,3 +1,24 @@
 # Block numbers and time
 
-TODO.
+On ZkSync there are two types of blocks: L2 blocks and L1 batches.
```
popzxc, line 3: Why do we need 2 types? Why the latter called "batches" and not "blocks" (though they're called blocks at the beginning of the sentence)?

docs/dev/zksync-v2/blocks-and-time.md
```diff
@@ -1,3 +1,24 @@
...
+
+## L2 blocks
+
+L2 block is similar to a common block on L1. L2 block time is almost constant, like a few seconds.
```
popzxc, line 7: What is a "common block"? Does "uncommon block" exist on L1?
"Almost constant, like a few seconds" sounds very uncertain -- is it a formal definition? Why can't it be "fully constant"?

docs/dev/zksync-v2/blocks-and-time.md
```diff
@@ -1,3 +1,24 @@
...
+In this documentation by default the word "block" actually means "L2 block" including API.
+L2 block hash is deterministic and is basically equal to `keccak256(bytes4(l2_block_number))`.
```
StanislavBreadless, line 9: Maybe let's have it as `keccak256(l2_block_number)`?

IMHO does not really matter, but the `bytes4` part complicates the thing a bit. Also, formally, there is a difference between 

```
keccak256(bytes.concat(bytes4(l2_block_number)))
```

and 

```
keccak256(bytes.concat(bytes32(uint256(l2_block_number))))
```

The former is left-aligned, while the latter is right-aligned. We don't need to elaborate that much in the documentation, but I am still curious :)

docs/dev/zksync-v2/blocks-and-time.md
```diff
@@ -1,3 +1,24 @@
 # Block numbers and time
 
-TODO.
+On ZkSync there are two types of blocks: L2 blocks and L1 batches.
+
+## L2 blocks
+
+L2 block is similar to a common block on L1. L2 block time is almost constant, like a few seconds.
+In this documentation by default the word "block" actually means "L2 block" including API.
```
popzxc, line 8: "word ... actually means" makes it sound like we wrote the documents incorrectly and are unable to use the correct term outselves.
To make it sound better we need to work on phrasing, i.e. "For simplicity, when we mention "block" in this documentation, "L2 block" is always assumed, unless specified otherwise".
"including API" part is a bit confusing to me -- what's so special about API?

docs/dev/zksync-v2/blocks-and-time.md
```diff
@@ -1,3 +1,24 @@
...
+Thus, L1 batch is sealed only if its capacity limit, or a timeout is reached.
+That is why L1 batch time can be different: from minutes under the high load to several hours when the batch is closed by timeout.
+
+Note, that it does not relate to L2 blocks, they are "cheap" and can be sealed fast.
+For users, monitoring L1 batches might be interesting only in case they are waiting for finality on L1,
+for all other cases they should use L2 blocks.
+
+## `block` properties
+While executing code block properties (`block.number`, `block.hash`, `block.timestamp`, etc) are actually the properties of the current L1 batch.
+So, you shouldn't rely on the fact that they are changing every few seconds.
```
popzxc, line 24: This is confusing and seemingly incorrect. If these are properties of L1 batch, you *can* rely on them, no?
You should not rely on the hash of L2 block, since it's deterministic. And there should be clarification of what does "rely" mean in this context.

docs/dev/zksync-v2/blocks-and-time.md
StanislavBreadless, line 8: I think this may is simpler:

```suggestion
In this documentation and zkSync API, the word "block" refers to the "L2 block".
```

docs/dev/zksync-v2/blocks-and-time.md
```diff
@@ -1,3 +1,24 @@
...
+## L2 blocks
+
+L2 block is similar to a common block on L1. L2 block time is almost constant, like a few seconds.
+In this documentation by default the word "block" actually means "L2 block" including API.
+L2 block hash is deterministic and is basically equal to `keccak256(bytes4(l2_block_number))`.
+
+## L1 batches
+
+L1 batch is a range of L2 blocks that are committed and finalized on L1 as one whole.
+The purpose of introducing L1 batch conception is to achieve cost optimization by using the whole capacity we have.
```
popzxc, line 14: What does "whole capacity" mean? What defines said capacity?

docs/dev/zksync-v2/blocks-and-time.md
```diff
@@ -1,3 +1,24 @@
...
+
+## L2 blocks
+
+L2 block is similar to a common block on L1. L2 block time is almost constant, like a few seconds.
+In this documentation by default the word "block" actually means "L2 block" including API.
+L2 block hash is deterministic and is basically equal to `keccak256(bytes4(l2_block_number))`.
+
+## L1 batches
+
+L1 batch is a range of L2 blocks that are committed and finalized on L1 as one whole.
```
StanislavBreadless, line 13: > committed and finalized

Just in case the users don't know what committing and finalising means, please add a link to the page, where we explain how ZK rollups work in general

docs/dev/zksync-v2/blocks-and-time.md
```diff
@@ -1,3 +1,24 @@
...
+The purpose of introducing L1 batch conception is to achieve cost optimization by using the whole capacity we have.
+Thus, L1 batch is sealed only if its capacity limit, or a timeout is reached.
```
popzxc, line 15: Why do we want to have a timeout if we want to utilize the whole capacity?

docs/dev/zksync-v2/blocks-and-time.md
```diff
@@ -1,3 +1,24 @@
...
+That is why L1 batch time can be different: from minutes under the high load to several hours when the batch is closed by timeout.
+
+Note, that it does not relate to L2 blocks, they are "cheap" and can be sealed fast.
+For users, monitoring L1 batches might be interesting only in case they are waiting for finality on L1,
+for all other cases they should use L2 blocks.
```
popzxc, line 20: How can they "use" L2 blocks? Is it opt-in?

docs/dev/zksync-v2/blocks-and-time.md
```diff
@@ -1,3 +1,24 @@
...
+
+## `block` properties
```
StanislavBreadless, line 22: There should always be a newline between a title and the content

```suggestion
## `block` properties

```

popzxc (commented):
Thanks for the update! However, I think that `blocks-and-time.md` can still be significantly improved. Both on readability and grammar/phrasing.

To demonstrate the idea, I tried to show questions users may have.

@krofax perhaps you could take care of it?

docs/dev/zksync-v2/blocks-and-time.md
```diff
@@ -1,3 +1,24 @@
...
+While executing code block properties (`block.number`, `block.hash`, `block.timestamp`, etc) are actually the properties of the current L1 batch.
```
StanislavBreadless, line 23:
```suggestion
While executing code block properties (`block.number`, `block.hash`, `block.timestamp`, etc) the properties of the current L1 batch are returned.
```

docs/dev/zksync-v2/blocks-and-time.md
StanislavBreadless, line 24:
```suggestion
That is why you shouldn't rely on the fact that they are changing every 10-15 seconds.
```

StanislavBreadless (commented):
Left some neats and proposals

barakshani (commented):
Turns out I didn't submit my review... there are some collisions with other reviews now. sorry

barakshani:
@perekopskiy maybe close this due to https://github.com/matter-labs/zksync-web-v2-docs/pull/97 ?

PR #114, opened 2022-10-15
Fix hardhat issues in hello-world.md

openluminus:
I ran into issues while following this tutorial.
I found out it's better to initialize a hardhat project then modify its config files. Also removed instructions on creating the 'contracts' folder, since hardhat does it.


uF4No:
For now the prefered approach is to install dependencies and create the `hardhat.confg.ts` and folders manually.
Both approaches are valid and require manually modifying files or creating folders so we'll try to improve how to scaffold a project

PR #120, opened 2022-10-17
Re-wrote edit docs page

krofax:
Updated and created the following pages;
1. updated edit-docs.md
2. Created docs.md


uF4No:
The "edit this page" links do not work. Looks like there is a double "/" in the path.

barakshani (approved):

PR #124, opened 2022-10-18
Block explorer

krofax:
### What I worked on
Added the documentation for everything block explorer which includes:

1. Blocks details on the explorer.
2. Transaction details on the explorer.
3. How to use the search bar.
4. Updated smart contract verification file.
5. Started working on zkEVM docs, but paused, because it's a work in progress on the FE team.


uF4No:
I think it's confusing that the "Contract Verification" navigation item in the Developer Docs now sends you to the Tools/SDKs > Block explorer. It would be difficult for a user that is not familiar with our docs to understand how to go back to the "Developer Docs".
What do you think if we create a simple page that mentions something like "You can verify the contracts deployed on zkSync using the official block explorer. All the details about how to verify contacts can be found here" (with a link that opens in a new tab). That way we keep users in the same docs section (Developer docs) and it's less confusing. What do you think?

docs/api/tools/block-explorer/block-view.md
```diff
@@ -0,0 +1,56 @@
...
+View all blocks for a more complete overview of recent blocks, or click on a specific block number for details on that block.
+The blocks page is also accessible via the top menu.
+
+![Access the block page](../../../assets/images/block-menu.png "Blocks menu")
+
+## View all blocks
+
+The view all blocks page shows updated details from the most recent blocks. Use the pagination to navigate to the next page.
+
+However, if you have already left the homepage, you can always search for any Ethereum, DAI, LINK, wBTC, and USDC block by typing in the block number.
```
uF4No, line 15: I don't think we need to mention tokens here. You can enter a number and you'll search for that specific block regardless of the tokens

uF4No (changes requested):

krofax:
> I think it's confusing that the "Contract Verification" navigation item in the Developer Docs now sends you to the Tools/SDKs > Block explorer. It would be difficult for a user that is not familiar with our docs to understand how to go back to the "Developer Docs". What do you think if we create a simple page that mentions something like "You can verify the contracts deployed on zkSync using the official block explorer. All the details about how to verify contacts can be found here" (with a link that opens in a new tab). That way we keep users in the same docs section (Developer docs) and it's less confusing. What do you think?

Totally agree with you.

uF4No (approved):
Approved

docs/api/tools/block-explorer/contract-verification.md
```diff
@@ -0,0 +1,88 @@
...
+  
+<br>
+
+![Compiler version!](../../../assets/images/compiler-version.png "compiler version")
+
+- Enter the Solidity Contract code: Copy the code from your editor and paste it into the text area.
+<br>
+**Note:** You may need to flatten your Solidity code if it utilizes a library or inherits dependencies from another contract. 
+<br> 
+We recommend using [Hardhat flatten](https://medium.com/coinmonks/flattening-smart-contracts-using-hardhat-dffe7dbc7b3f), [Truffle flattener](https://github.com/NomicFoundation/truffle-flattener) or the [POA Solidity flattener](https://github.com/poanetwork/solidity-flattener).
```
uF4No, line 56: Added recommend `hardhat flatten` first

uF4No (approved):
Approved

uF4No (approved):
looks good

PR #155, opened 2022-11-01
change funds amount on deployment

nicobevilacqua:
Calculate the needed gas amount to avoid insufficient balance error on deployment


uF4No:
Good idea to dynamically check the cost of deployment as we do in other tutorials. Thanks for the update @nicobevilacqua 🤙

PR #160, opened 2022-11-07
fix: Updated Metamask page

krofax:
Added a link to block explorer and specified that users can add a network manually.


uF4No (approved):
approved

PR #180, opened 2022-11-23
Added an FAQ

krofax:
## Description
Stated that an address cannot be claimable by an adversary in zkSync 2.0.


docs/dev/fundamentals/faq.md
```diff
@@ -158,9 +158,13 @@ The transactions on zkSync 2.0 should not take more than 5 minutes.
 
 Our [Block Explorer](https://explorer.zksync.io) will show everything you may need about a transaction. 
 
+## Will an address be claimable by an adversary in zkSync 2.0?
+
+No! We have a different address generation schema from other L2's, that would not allow addresses to be claimed on L2 by an adversary. Even if you deploy same address and same private key it would still be different.
```
githubdoramon, line 163: doesn't make sense to say "if you deploy same address" given the address is not deployed, but a contract. Do we mean "if you deploy the exact same contract, with same private key and nouce"?

Also, could we maybe be more descriptive on this situation? Maybe instead of calling it adversary, generalizing like "can someone claim the address I have for my contract in other EVM networks, etc etc etc"? 

It is also important to make it clear then that you cannot have the same address as you have in other networks right?

githubdoramon (approved):

PR #182, opened 2022-11-23
Updated network details info

krofax:
Updated network details info across all affected pages.


githubdoramon (approved):
Although this is good, I'll create another ticket so we consolidate this info in a single place in the docs, and add links to it

PR #185, opened 2022-11-25
Updated content on contract verification page

krofax:
Updated content on contract verification page


obmatter (approved):
Review from my side:
The issue with the wrong number of space bars -> fixed.
Additional information about sol version is updated.

PR #201, opened 2022-12-01
Added TOC to docs

krofax:
Added table of content to every page in the docs.


uF4No:
I've removed the borders as they were unnecessary and it looked bad with the page footer

githubdoramon (approved):

PR #223, opened 2022-12-15
update event page

krofax:
Updated description on fetching events and added sample code in JS.


docs/dev/developer-guides/building-on-zksync/events.md
```diff
@@ -14,30 +14,50 @@ At zkSync, events behave the same way as in Ethereum.
...
-    abi = [
-  "event Transfer(address indexed src, address indexed dst, uint val)"
-];
+```js
+const ether = require("ethers");
+const contractABI = require("./ABI_JSON")
 
-contract = new Contract(tokenAddress, abi, provider);
+const getEvent = async () => {
+const usdtAdress = "<YOUR ADDRESS>"
```
githubdoramon, line 30: This code block looks weird... mentions YOUR ADDRESS for a usdt contract for example, from and to is the same block number, has an "event" parameter that doesn't say what this variable actually is, mentions PAST_EVENT that is nowhere in the code.

Needs some work in order to be a good code example

githubdoramon (changes requested):

docs/dev/developer-guides/building-on-zksync/events.md
krofax: Did a rework on the code sample. @githubdoramon

docs/dev/developer-guides/building-on-zksync/events.md
```diff
@@ -14,30 +14,50 @@ At zkSync, events behave the same way as in Ethereum.
 ## Events filtering
 
 Filtering is used to query indexed data and provide lower-cost data storage when the data is not required to be accessed on-chain.
+While filtering, we advice to load events by block ranges (0-1999, 2000-3999, ...) and index the result on your end, else you will get an error that says "block range should be less than or equal to 2000".
```
githubdoramon, line 17: When filtering, you should load events by block ranges (0-1999, 2000-3999, ...) and index the result on your end, or else you will get an error that says "block range should be less than or equal to 2000".

docs/dev/developer-guides/building-on-zksync/events.md
```diff
@@ -14,30 +14,50 @@ At zkSync, events behave the same way as in Ethereum.
...
-    abi = [
-  "event Transfer(address indexed src, address indexed dst, uint val)"
-];
+```js
+const ether = require("ethers");
+const contractABI = require("./ABI_JSON")
 
-contract = new Contract(tokenAddress, abi, provider);
+const getEvent = async () => {
+const contractAddress = "<YOUR ADDRESS>"
```
githubdoramon, line 30: < CONTRACT ADDRESS >

docs/dev/developer-guides/building-on-zksync/events.md
```diff
@@ -14,30 +14,50 @@ At zkSync, events behave the same way as in Ethereum.
...
-//        "toBlock": "0x8e8e6",
-//        "topics": [["0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef"]]
-//    }]
-//}
+const contract = new ether.ethers.Contract(contractAddress, contractABI, provider);
+
+await contract.on("Transfer", (event) => {
+    let info = {
+        filter: { INDEXED_PARAMETER: VALUE },
+        fromBlock: BLOCK_NUMBER, // e.g 15943000
```
githubdoramon, line 38: < START BLOCK NUMBER >

docs/dev/developer-guides/building-on-zksync/events.md
```diff
@@ -14,30 +14,50 @@ At zkSync, events behave the same way as in Ethereum.
...
+        toBlock: BLOCK_NUMBER, // e.g 15943100
```
githubdoramon, line 39: < END BLOCK NUMBER >

docs/dev/developer-guides/building-on-zksync/events.md
```diff
@@ -14,30 +14,50 @@ At zkSync, events behave the same way as in Ethereum.
...
-//       "fromBlock": "0x8e8e6",
-//        "toBlock": "0x8e8e6",
-//        "topics": [["0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef"]]
-//    }]
-//}
+const contract = new ether.ethers.Contract(contractAddress, contractABI, provider);
+
+await contract.on("Transfer", (event) => {
+    let info = {
+        filter: { INDEXED_PARAMETER: VALUE },
```
githubdoramon, line 37: What are those constants?

docs/dev/developer-guides/building-on-zksync/events.md
```diff
@@ -14,30 +14,50 @@ At zkSync, events behave the same way as in Ethereum.
...
+getEvent()
 
 ```
-zkSync has a 10K log limit per response.
+
+where
+
+Provider — Your websocket provider through which you will retrieve the events data.
+Contract address — The contract address whose events you want to track.
+ABI — The ABI(Application Binary Interface) of the contract in JSON format.
```
githubdoramon, line 54: missing space before parenthesis

docs/dev/developer-guides/building-on-zksync/events.md
```diff
@@ -14,30 +14,50 @@ At zkSync, events behave the same way as in Ethereum.
...
+Transfer — Any name you want to assign to print it at the end, for this sample, we used "Transfer".
+Event name — The name of the event as defined in the smart contract.
```
githubdoramon, line 56: Where is event name in the code?

docs/dev/developer-guides/building-on-zksync/events.md
```diff
@@ -14,30 +14,50 @@ At zkSync, events behave the same way as in Ethereum.
...
+Indexed parameter — The indexed parameter of the event.
```
githubdoramon, line 57: These 2 descriptions (Indexed parameter and Value) are not really helpfull

githubdoramon (changes requested):

docs/dev/developer-guides/building-on-zksync/events.md
uF4No: fixed in 79e18fb8c9dd426d700ab809eaeff8251ccb799e

docs/dev/developer-guides/building-on-zksync/events.md
uF4No: fixed in 79e18fb8c9dd426d700ab809eaeff8251ccb799e

docs/dev/developer-guides/building-on-zksync/events.md
uF4No: I this is fixed in 79e18fb8c9dd426d700ab809eaeff8251ccb799e

docs/dev/developer-guides/building-on-zksync/events.md
```diff
@@ -14,30 +14,49 @@ At zkSync, events behave the same way as in Ethereum.
...
+
+
+- Provider: Your websocket provider through which you will retrieve the events data. Note that you need to use the websocket endpoint.
+- Contract address: The contract address whose events you want to track.
+- ABI: The ABI (Application Binary Interface) of the contract in JSON format.
+- Event name: The name of the event as defined in the smart contract. In this example we used the "Transfer" event from an ERC20 contract.
+- Indexed parameter (optional) : The indexed parameter of the event.
+- Block number (optional) The block number range for the events retrieval.
+
+**Note**: zkSync has a 10K log limit per response, however, if you receive a response with 10k events then there are more than 10k events in the block range, otherwise, there are less than 10k events.
```
uF4No, line 60: I think this phrase is a little confusing. What do you think about this:

> zkSync has a 10K log limit per response. This means that if you receive a response with 10k events, it will most likely contain additional events so it'd be a good idea to adjust the filters to retrieve the events in multiple batches.

krofax: What part do you think will confuse a reader?

krofax: @githubdoramon what do you suggest?

uF4No: I mean the note about the 10k log limit

githubdoramon: better indeed

docs/dev/developer-guides/building-on-zksync/events.md
```diff
@@ -14,30 +14,49 @@ At zkSync, events behave the same way as in Ethereum.
...
+listenEvents()
 
 ```
-zkSync has a 10K log limit per response.
+
+
+- Provider: Your websocket provider through which you will retrieve the events data. Note that you need to use the websocket endpoint.
+- Contract address: The contract address whose events you want to track.
+- ABI: The ABI (Application Binary Interface) of the contract in JSON format.
+- Event name: The name of the event as defined in the smart contract. In this example we used the "Transfer" event from an ERC20 contract.
```
githubdoramon, line 56: we are not referring to event name in the code... maybe we should add as a constant to this explanation makes more sense

docs/dev/developer-guides/building-on-zksync/events.md
```diff
@@ -14,30 +14,49 @@ At zkSync, events behave the same way as in Ethereum.
...
+- Indexed parameter (optional) : The indexed parameter of the event.
```
githubdoramon, line 57: Can we provide an example of what exactly is an indexed parameter? Maybe I am too noob, but I have no idea

docs/dev/developer-guides/building-on-zksync/events.md
```diff
@@ -14,30 +14,49 @@ At zkSync, events behave the same way as in Ethereum.
...
+- Block number (optional) The block number range for the events retrieval.
```
githubdoramon, line 58: Let's refer to start and end block number. And it is not really optional, after all we are stating in the beginning that if dev dones't use it, it will get an error. Would be better to explain this here

githubdoramon (changes requested):

docs/dev/developer-guides/building-on-zksync/events.md
krofax: The index parameter is where you specify the attribute name and identify the field or fields on which the index is based.
E.g - from: '0x48c6c0923b514db081782271355e5745c49wd60'

githubdoramon: Ah.. doh, stupid me.. ok... so worth adding an example like the one you did

docs/dev/developer-guides/building-on-zksync/events.md
githubdoramon: We should use Antonio's proposal here

uF4No: @krofax just missing this. Can you replace the last paragraph with what's mentioned above?

uF4No:
I think we should point the Python SDK menu item directly to the "Getting started" page, and add the paragraph currently in the Python/Readme, inside the "Getting started" page as well:

> While most of the existing SDKs should work out of the box, deploying smart contracts or using unique zkSync features, like paying fees in other tokens, requires providing additional fields to those that Ethereum transactions have by default.
> To provide easy access to all of the features of zkSync 2.0, the zksync2 Python SDK was created.

krofax:
> While most of the existing SDKs should work out of the box, deploying smart contracts or using unique zkSync features, like paying fees in other tokens,

That's how it is in the Python PR.

uF4No (approved):

PR #224, opened 2022-12-18
Add hardhat-abi-exporter

mpopovac-txfusion:


uF4No:
I think that we should create different pages for zksync-specifc plugins and plugins that work out of the box like this one. Just because if we keep adding more plugins here it'll be a super long list and I don't think mixing them would be a good idea.
What do you think @krofax @githubdoramon ?

githubdoramon:
> I think that we should create different pages for zksync-specifc plugins and plugins that work out of the box like this one. Just because if we keep adding more plugins here it'll be a super long list and I don't think mixing them would be a good idea. What do you think @krofax @githubdoramon ?

Agree... makes a lot of sense to have a separate section for "regular" plugins that can be used out of the box

krofax:
> I think that we should create different pages for zksync-specifc plugins and plugins that work out of the box like this one. Just because if we keep adding more plugins here it'll be a super long list and I don't think mixing them would be a good idea. What do you think @krofax @githubdoramon ?

Totally agree too, it's almost already looking long.

docs/api/hardhat/plugins.md
```diff
@@ -411,3 +411,36 @@ expect(await contract.getAmount()).to.equal(100);
 ```
 
 Checkout the advantages of using chai matchers [here](https://hardhat.org/hardhat-chai-matchers/docs/overview#why-would-i-want-to-use-it?). Since the list of all supported chai matchers is same as with [hardhat-chai-matchers](https://hardhat.org/hardhat-chai-matchers/docs/overview) plugin, check the [reference documentation](https://hardhat.org/hardhat-chai-matchers/docs/reference).
+
+## `hardhat-abi-exporter`
+
+This plugin allows developers to generate the ABI files for given smart contracts easily. It saves them in a specified location, allowing developers to access and use the ABIs in their projects.
+
+> **_NOTE:_** The hardhat-abi-exporter plugin does not require modifications because there is no need to override tasks or subtasks. As a result, it can be used as is. For more information, you can check the plugin's [GitHub page](https://github.com/ItsNickBarry/hardhat-abi-exporter).
```
krofax, line 419: It can be used as it is.

uF4No:
we should create a page for plugins that work out-of-the-box and just link to their official docs.

PR #253, opened 2023-01-16
adding Swift basic info

githubdoramon:


docs/api/swift/getting-started.md
```diff
@@ -0,0 +1,173 @@
+# Getting started
+
+In this guide we will demonstrate how to:
+
+1. Connect to the zkSync network.
```
githubdoramon, line 5: Add the other things that we will "tech" here @krofax

krofax (approved):

PR #263, opened 2023-01-30
Separated the hardhat plugins

krofax:
Separated the hardhat plugins


uF4No:
If we're going to leave the zksync-toolbox as a main page for plugins, I'd rename it to "zkSync toolbox" so it appears with that name in the menu. It's weird to have a "Plugins" section in the menu, but then have all the plugins outside of it 🫤

krofax:
> If we're going to leave the zksync-toolbox as a main page for plugins, I'd rename it to "zkSync toolbox" so it appears with that name in the menu. It's weird to have a "Plugins" section in the menu, but then have all the plugins outside of it 🫤

Super valid 💯

uF4No (approved):

PR #264, opened 2023-01-30
Multi file verification update

FilipTxFusion:


uF4No:
@FilipTxFusion  We've made some changes in the content organisation and now each plugin has its own page inside `/docs/api/hardhat/`, so you'd need to add this new paragraphs in the hardhat verify plugin page instead of the general plugins page.

FilipTxFusion:
> @FilipTxFusion We've made some changes in the content organisation and now each plugin has its own page inside `/docs/api/hardhat/`, so you'd need to add this new paragraphs in the hardhat verify plugin page instead of the general plugins page.

Modified to match the newest structure.

uF4No (approved):

uF4No:
approved

PR #268, opened 2023-02-02
Added warning messages for ecrecover and tx.origin respectively

krofax:
Added warning messages for ecrecover and tx.origin respectively.


uF4No:
I think this includes the upgrade to Vuepress v2, so it should not target the main branch yet.

uF4No:
I've changed the format to a section named "Other considerations" to not have just warning messages without any context. We could add info about the timestamp and block number here as well in the upcoming days. For now, It looks good to me.

uF4No (approved):

PR #275, opened 2023-02-07
[WIP]Update all doc after M5 regenesis on stage2

AnastasiiaVashchuk:


docs/api/js/accounts-l1-l2.md
```diff
@@ -179,27 +180,37 @@ async claimFailedDeposit(depositHash: BytesLike): Promise<ethers.ContractTransac
...
+| transaction.operatorTip (optional) | If the ETH `value` passed with the transaction is not explicitly stated in the overrides, this field will be equal to the tip the operator will receive on top of the base cost of the transaction. This value has no meaning for the `Deque` type of queue, but it will be used to prioritize the transactions that get into the `Heap` or `HeapBuffer` queues. 
+                                                                                                        |
+| overrides (optional)               | Ethereum transaction overrides. May be used to pass `gasLimit`, `gasPrice` etc.
+                                |
+| transaction.gasPerPubdataByte (optional)      |
+                              | 
+| transaction.refundRecipient(optional)         |  
+|--                                                                                                                                                         
+| returns                               | `PriorityOpResponse` object.
```
AnastasiiaVashchuk, line 213: @krofax  I updated the context of the table, but it is displayed very bed in the preview for me...could you please correct it

docs/api/js/accounts-l1-l2.md
```diff
@@ -228,20 +239,21 @@ const abi = [
...
 });
 
 console.log(`Executing the transaction will cost ${ethers.utils.formatEther(txCostPrice)} ETH`);
 
 const executeTx = await wallet.requestExecute({
-  calldata,
-  l2gasLimit,
   contractAddress: "0x19a5bfcbe15f98aa073b9f81b58466521479df8d",
+  calldata,
+  l2Value: 1,
```
StanislavBreadless, line 255: why do we supply 1 here?

docs/dev/developer-guides/bridging/l1-l2.md
```diff
@@ -7,7 +7,7 @@ This section describes the interface for interaction with zkSync from L1. It ass
 
 ::: warning
 
-Please note that in the new `0.13.0-beta.2` SDK we changed the API to rename ergs to gas, to avoid breaking changes in your integration, update all your project dependencies.
+Please note that in the new `0.13.0-beta.2` SDK the API layer operates with gas. The ergs consept is used by VM only.
```
StanislavBreadless, line 10: Please change it in all other places too

```suggestion
Please note that in the new `0.13.0-beta.2` SDK the API layer operates with gas. The ergs concept is used by VM only.
```

uF4No: fixed

docs/api/js/accounts-l1-l2.md
uF4No: fixed

docs/api/js/accounts-l1-l2.md
krofax: @AnastasiiaVashchuk any reason why?

krofax (changes requested):

docs/api/js/accounts-l1-l2.md
```diff
@@ -73,17 +73,17 @@ async deposit(transaction: {
...
-| transaction.amount                      | The amount of the token to be deposited.                                                                                                                                                                                                                                                                                                                         |
-| transaction.to (optional)               | The address that will receive the deposited tokens on L2.                                                                                                                                                                                                                                                                                                        |                                       
+| Name                                    | Description      |
+| --------------------------------------- | ------- |
+| transaction.token                       | The address of the token to deposit  |
```
arose0, line 78: nitpick: missing period after "deposit".

docs/api/js/accounts-l1-l2.md
```diff
@@ -153,7 +153,7 @@ async getBaseCost(params: {
 
 | Name                        | Description                                                                                            |
 | --------------------------- | ------------------------------------------------------------------------------------------------------ |
-| params.l2gasLimit           | The `l2gasLimit` for the call.                                                                          |
+| params.gasLimit             | The `gasLimit` for the call.                                                                          |
```
arose0, line 156: Just a suggestion, but maybe it is worth making it clear when we're talking about L1 and L2 here. I note below that we still say "the gas price of the L1 transaction" and we should make sure we're being as unambiguous as possible here.

docs/api/js/accounts-l1-l2.md
AnastasiiaVashchuk: @StanislavBreadless  randomly, or I understand the mission of this variable wrong?

AnastasiiaVashchuk: @StanislavBreadless  there is also contract address value... For me looks random - there is no explanation for what is it

StanislavBreadless: @AnastasiiaVashchuk `l2Value` is a field which denotes how much ETH to you want to be present within this L1->L2 transaction

AnastasiiaVashchuk: so if it is an example why we can not just pass `l2Value` = 1?

docs/api/js/accounts-l1-l2.md
AnastasiiaVashchuk: absolutely agree!
will fix.

docs/api/js/accounts-l1-l2.md
StanislavBreadless: @AnastasiiaVashchuk we can, but then we should explain what we are doing here. It is just in the previous example no value was passed

krofax (approved):

PR #318, opened 2023-02-28
Gas estimation for transaction with custom and paymasters accounts

krofax:
PR contains the following:

- Gas estimation for transactions with custom and paymaster accounts.
- Code tabs across all tutorials.

### DIFF
--- a/docs/api/hardhat/plugins.md
+++ b/docs/api/hardhat/plugins.md
@@ -16,13 +16,22 @@ List of supported plugins:
 
 Add the latest version of this plugin to your project with the following command:
 
-```
-# Yarn
-yarn add -D @matterlabs/hardhat-zksync-toolbox @matterlabs/hardhat-zksync-solc @matterlabs/hardhat-zksync-vyper @matterlabs/hardhat-zksync-chai-matchers @matterlabs/hardhat-zksync-deploy @matterlabs/hardhat-zksync-verify hardhat ethers zksync-web3 @nomicfoundation/hardhat-chai-matchers @nomiclabs/hardhat-ethers @nomiclabs/hardhat-etherscan
+::: code-tabs
 
-# Npm (version 7 or later is recommended)
+@tab:active yarn
+
+```bash
+yarn add -D @matterlabs/hardhat-zksync-toolbox @matterlabs/hardhat-zksync-solc @matterlabs/hardhat-zksync-vyper @matterlabs/hardhat-zksync-chai-matchers @matterlabs/hardhat-zksync-deploy @matterlabs/hardhat-zksync-verify hardhat ethers zksync-web3 @nomicfoundation/hardhat-chai-matchers @nomiclabs/hardhat-ethers @nomiclabs/hardhat-etherscan
+```
+
+@tab npm
+
+```bash
 npm i -D @matterlabs/hardhat-zksync-toolbox
 ```
+:::
+
+> Note: For `npm`, version 7 or later is recommended.
 
 ### Usage
 
--- a/docs/dev/building-on-zksync/hello-world.md
+++ b/docs/dev/building-on-zksync/hello-world.md
@@ -28,18 +28,29 @@ Please note that breaking changes were introduced in `zksync-web3 ^0.13.0`. The 
 
 1. Initialize the project and install the dependencies. Run the following commands in your terminal:
 
+::: code-tabs
+
+@tab:active mkdir
+
 ```sh
 mkdir greeter-example
 cd greeter-example
+```
 
-# For Yarn
+@tab yarn
+
+```bash
 yarn init -y
 yarn add -D typescript ts-node ethers@^5.7.2 zksync-web3@^0.13.1 hardhat @matterlabs/hardhat-zksync-solc @matterlabs/hardhat-zksync-deploy
+```
 
-# For NPM
+@tab npm
+
+```bash
 npm init -y
 npm i -D typescript ts-node ethers@^5.7.2 zksync-web3@^0.13.1 hardhat @matterlabs/hardhat-zksync-solc @matterlabs/hardhat-zksync-deploy
 ```
+:::
 
 Please note that Typescript is required by zkSync plugins.
 
@@ -196,16 +207,28 @@ git clone https://github.com/matter-labs/greeter-tutorial-starter
 
 2. Spin up the project:
 
+::: code-tabs
+
+@tab:active cd
+
 ```sh
 cd greeter-tutorial-starter
-# For Yarn
+```
+
+@tab yarn
+
+```bash
 yarn
 yarn serve
+```
 
-# For NPM
+@tab npm
+
+```bash
 npm install
 npm run serve
 ```
+:::
 
 By default, the page should be running at `http://localhost:8080`. Open this URL in the browser to see the page.
 
@@ -283,13 +306,20 @@ const GREETER_CONTRACT_ABI = []; // TODO: Complete and import the ABI
 
 Run the following command on the greeter-tutorial-starter root folder to install `zksync-web3` and `ethers`:
 
-```
-# For Yarn
-yarn add ethers@^5.7.2 zksync-web3@^0.13.1
+::: code-tabs
 
-# For NPM
+@tab:active yarn
+
+```bash
+yarn add ethers@^5.7.2 zksync-web3@^0.13.1
+```
+
+@tab npm
+
+```bash
 npm i ethers@^5.7.2 zksync-web3@^0.13.1
 ```
+:::
 
 After that, import both libraries in the `script` part of the `App.vue` file (right before the contract constant). It should look like this:
 
@@ -542,7 +572,7 @@ Even though ether is the only token you can pay fees with, the account abstracti
 
 ::: warning Paymasters on mainnet
 
-🚨 The testnet paymaster is purely for demonstration of this feature and won't be available on mainnet. When integrating your protocol on mainnet, you should follow the documentation of the paymaster you will use or create your own.
+The testnet paymaster is purely for demonstration of this feature and won't be available on mainnet. When integrating your protocol on mainnet, you should follow the documentation of the paymaster you will use or create your own.
 
 :::
 
--- a/docs/dev/developer-guides/transactions/fee-model.md
+++ b/docs/dev/developer-guides/transactions/fee-model.md
@@ -18,9 +18,6 @@ Since the costs for publishing the calldata on L1 are very volatile, the number 
 
 **Please note that the public data is published only for state diffs.** If the same storage slot is updated 10 times in the same rollup block, only the final update will be published on Ethereum, thus only charging for public data once.
 
-
-
-
 ## Fee model: zkSync vs Ethereum
 
 We want to show the clear distinction between our fee model and the Ethereum one.
@@ -39,3 +36,67 @@ So the tips to make the most out of the zkSync fee system are the following:
 - **Reuse as many storage slots as possible.** Only the state diff is published on Ethereum.
 - **Users should share as many storage slots as possible.** If 100 users update a storage slot of your contract in a single block, the diff will be published only once. In the future, we will introduce reimbursement for the users, so that the costs for updating shared storage slots are split between the users.
 - **Reuse the contract code if possible.** On Ethereum, avoiding constructor parameters and putting them into constants reduces some of the gas costs upon contract deployment. On zkSync the opposite is true: deploying the same bytecode for contracts, while changing only constructor parameters can lead to substantial fee savings.
+
+
+## Gas estimation during transaction: For custom and paymaster accounts
+
+On Ethereum, there is a constant of 21000 gas that should cover all the intrinsic costs of processing a transaction: checking the signature and updating the nonce of the account. 
+
+On zkSync Era, it's quite different, because we support custom accounts that may take a slightly different (typically a bit higher) number of gas than EOAs. 
+We provide a good, robust way for anyone to estimate the cost of a transaction regardless of the type of account. 
+### Changes in the `validateTransaction`
+
+Right now, the `validateTransaction` fails if the signature is not correct and returns `success=true` otherwise, while the `returndata` is discarded completely. We will migrate to a different approach, where the `validateTransaction` method to be considered successful must return some magic string. Any other `returndata` will be considered invalid and will be rejected by the system.  
+
+Now, `validateTransaction` method of the AA (or the paymaster `validateAndPayForPaymasterTransaction` method), should always try to do the same amount of computation (and especially storage accesses) as it would have done if the transaction was indeed validated correctly. By default, the operator will provide a transaction structure that contains as much information as is available to the operator at the time of the fee estimation. As a substitute for the signature, an invalid 65-byte ECDSA signature will be used. The `DefaultAccount` (the one that is used by EOAs) will do as many operations as possible (including signature verification) and only return `bytes4(0)` instead of magic. If, for instance, the wallet is a custom account, with multiple signers, then it may want to simulate the validation of the signatures of all the provided signers.
+
+So roughly the code of the validation step of each account will look the following way:
+
+```solidity
+ function validateTransaction(
+        bytes32, // _txHash
+        bytes32 _suggestedSignedHash,
+        Transaction calldata _transaction
+    ) external payable override ignoreNonBootloader returns (bytes4 magic) {
+        // ...
+        // normal validation logic
+        // ...
+         
+        // The execution before the following `if` statement should look as close as to the execution under incorrect 
+        // signature as possible
+        if(!validSignature) {
+           // Setting magic to 0 to make sure that the transaction will not be accepted on mainnet
+           magic = bytes4(0);  
+        } else {
+           magic = // setting to actual magic value which should be returned by the correct signature validation
+        }
+    }
+```
+
+### Notes on the transaction’s length
+
+zkSync Era sends state diffs onchain, and the cost for the transaction will still mildly depend on its length (because long transactions need to be stored in memory of the operator), also long transactions incur additional costs during interactions with an account. However, the signature (as well as its length) is not available at the time of fee estimation and so there is no correct way to precisely estimate the cost of the transaction. For now, we will compensate for it by multiplying the recommended cost of the transaction by **5%**. In the future, we will introduce the following:
+
+- Each account will be able to implement a method called `fillPartialTransaction` that will fill the signature with the substituted value which will be used for fee estimation.
+
+### Changes for `eth_estimateGas`
+
+Since the entire flow for tx validation + execution will be executed to get the transaction’s fee, the operator will manually put the necessary balance (if needed) for the account before the emulation is run to conduct the default account’s `payForTransaction` method.
+
+Note, that the operator will NOT do the same with the paymaster (i.e. if a transaction has a paymaster, the paymaster MUST have the proper balance beforehand). 
+
+The `eth_estimateGas` method itself will use binary search to find the smallest gas value under which the transaction *succeeds*.
+
+### Difference from geth
+
+Just like geth, we will use binary search. However, there will be some notable differences in gas estimation from the behavior of geth:
+
+- Unlike geth, it is impossible to track out of gas errors on zkSync Era. The main reason is that the “actual” execution will happen inside the DefaultAccount system contract and due to the 63/64 rule when a high number of gas is provided, the call to the `execute` method of the DefaultAccount will NOT fail even though the subcall to the `transaction.to` contract did fail with an out of gas error.
+
+That means that while geth can distinguish between “empty error” and “out of gas”, we can’t. If during gas estimation we fail to find the necessary `gasLimit`, will return the following error _“can not estimate gas: the transaction either fails or requires more gas than allowed”_.
+
+- During simulation, geth uses `tx.gasprice = 0` to make sure that the user can pay the fee even though the `tx.origin` in the simulation may not have any balance at all - This means that when `estimateGas` from an empty account is called, no `value` can be provided to such call as this account has zero balance to cover this value. 
+
+We could do that, but that would mean that the `payForTransaction` of the Account Abstraction protocol would do nothing and thus be much cheaper than it will be during the actual transaction validation. Instead, the operator will increase the balance of the user’s account by `tx.maxFeePerErg * tx.ergsLimit`.
+
+For DefaultAccount it will behave the same way as on geth (since the user will get rid of the new funds in the `payForTransaction` method), but for custom accounts, they may unexpectedly contain more balance than they will have onchain, which may affect their behavior.
--- a/docs/dev/tutorials/cross-chain-tutorial.md
+++ b/docs/dev/tutorials/cross-chain-tutorial.md
@@ -155,14 +155,20 @@ main().catch((error) => {
 
 5. Compile the contract and run the deployment script with:
 
-```
-# compile contract
+::: code-tabs
+
+@tab:active Compile contract
+
+```bash
 yarn hardhat compile
-
-# deploy contract
-yarn hardhat run --network goerli ./scripts/deploy.ts
-
 ```
+
+@tab Deploy contract
+
+```bash
+yarn hardhat run --network goerli ./scripts/deploy.ts
+```
+:::
 
 The last command will print the deployed governance smart contract address in the terminal.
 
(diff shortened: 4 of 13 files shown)


docs/dev/developer-guides/transactions/fee-model.md
```diff
@@ -39,3 +36,67 @@ So the tips to make the most out of the zkSync fee system are the following:
...
 - **Reuse the contract code if possible.** On Ethereum, avoiding constructor parameters and putting them into constants reduces some of the gas costs upon contract deployment. On zkSync the opposite is true: deploying the same bytecode for contracts, while changing only constructor parameters can lead to substantial fee savings.
+
+
+## Gas estimation during transaction: For custom and paymaster accounts
+
+On Ethereum, there is a constant of 21000 gas that should cover all the intrinsic costs of processing a transaction: checking the signature and updating the nonce of the account. 
+
+On zkSync Era, it's quite different, because we support custom accounts that may take a slightly different (typically a bit higher) number of gas than EOAs. 
+We provide a convenient way for anyone to estimate the cost of a transaction regardless of the type of account. 
+### Changes in the `validateTransaction`
```
StanislavBreadless, line 47: There should be a space before the title

```suggestion

### Changes in the `validateTransaction`
```

docs/dev/developer-guides/transactions/fee-model.md
```diff
@@ -39,3 +36,67 @@ So the tips to make the most out of the zkSync fee system are the following:
...
+
+Right now, the `validateTransaction` fails if the signature is not correct and returns `success=true` otherwise, while the `returndata` is discarded completely. We will migrate to a different approach, where the `validateTransaction` method to be considered successful must return some magic string. Any other `returndata` will be considered invalid and will be rejected by the system.
```
StanislavBreadless, line 49: This was a part of the internal spec. Please remove this paragraph
```suggestion
It used to be the case for `validateTransaction` to fail if the signature is not correct and return `success=true` otherwise, while the `returndata` is discarded completely. We have migrated to a different approach, where the `validateTransaction` method to be considered successful must return a magic string. Any other `returndata` will be considered invalid and will be rejected by the system.  
```

docs/dev/developer-guides/transactions/fee-model.md
```diff
@@ -39,3 +36,67 @@ So the tips to make the most out of the zkSync fee system are the following:
...
+
+Now, `validateTransaction` method of the AA (or the paymaster `validateAndPayForPaymasterTransaction` method), should always try to do the same amount of computation (and especially storage accesses) as it would have done if the transaction was indeed validated correctly. By default, the operator will provide a transaction structure that contains as much information as is available to the operator at the time of the fee estimation. As a substitute for the signature, an invalid 65-byte ECDSA signature will be used. The `DefaultAccount` (the one that is used by EOAs) will do as many operations as possible (including signature verification) and only return `bytes4(0)` instead of magic. If, for instance, the wallet is a custom account, with multiple signers, then it may want to simulate the validation of the signatures of all the provided signers.
```
StanislavBreadless, line 51: please use Present Simple when talking about the current properties of the system

docs/dev/developer-guides/transactions/fee-model.md
```diff
@@ -39,3 +36,67 @@ So the tips to make the most out of the zkSync fee system are the following:
...
+
+So roughly the code of the validation step of each account will look the following way:
```
StanislavBreadless, line 53: Please use this section as a guide for external developers on how they should build their custom accounts. The doc that I've sent to you was about how we built DefaultAccount (EOAs).

At the very least the we should rather use `should ` instead of `will`, etc. Generally we should use Present Simple here (`do/does`) or recommendations (`should`)

docs/dev/developer-guides/transactions/fee-model.md
```diff
@@ -39,3 +36,67 @@ So the tips to make the most out of the zkSync fee system are the following:
...
+
+```solidity
+ function validateTransaction(
```
StanislavBreadless, line 56: It is better to supply the link to the DefaultAccount implementation here: 

https://github.com/matter-labs/era-system-contracts/blob/5a6c728576de5db68ad577a09f34e7b85c374192/contracts/DefaultAccount.sol#L65

docs/dev/developer-guides/transactions/fee-model.md
```diff
@@ -39,3 +36,67 @@ So the tips to make the most out of the zkSync fee system are the following:
...
+           magic = bytes4(0);  
+        } else {
+           magic = // setting to actual magic value which should be returned by the correct signature validation
+        }
+    }
+```
+
+### Notes on the transaction’s length
+
+zkSync Era sends state diffs onchain, and the cost for the transaction will still mildly depend on its length (because long transactions need to be stored in memory of the operator), also long transactions incur additional costs during interactions with an account. However, the signature (as well as its length) is not available at the time of fee estimation and so there is no correct way to precisely estimate the cost of the transaction. For now, we will compensate for it by multiplying the recommended cost of the transaction by **5%**. In the future, we will introduce the following:
```
StanislavBreadless, line 78: Let's not hardcode "5%" here. It is better to give us some room to change it without updating the documentation. Better to say "by a few percents".

Also, it is better not to say "we will introduce the following" (I am sorry if it was present in the doc this way). We should rather say "may", because we do not commit to it, though it is likely what may happen in the future

docs/dev/developer-guides/transactions/fee-model.md
```diff
@@ -39,3 +36,67 @@ So the tips to make the most out of the zkSync fee system are the following:
...
+
+- Each account will be able to implement a method called `fillPartialTransaction` that will fill the signature with the substituted value which will be used for fee estimation.
+
+### Changes for `eth_estimateGas`
+
+Since the entire flow for tx validation + execution will be executed to get the transaction’s fee, the operator will manually put the necessary balance (if needed) for the account before the emulation is run to conduct the default account’s `payForTransaction` method.
+
+Note, that the operator will NOT do the same with the paymaster (i.e. if a transaction has a paymaster, the paymaster MUST have the proper balance beforehand).
```
StanislavBreadless, line 86:
```suggestion
```

It was an outdated part, it does

docs/dev/developer-guides/transactions/fee-model.md
```diff
@@ -39,3 +36,67 @@ So the tips to make the most out of the zkSync fee system are the following:
...
+
+Note, that the operator will NOT do the same with the paymaster (i.e. if a transaction has a paymaster, the paymaster MUST have the proper balance beforehand). 
+
+The `eth_estimateGas` method itself will use binary search to find the smallest gas value under which the transaction *succeeds*.
+
+### Difference from geth
+
+Just like geth, we will use binary search. However, there will be some notable differences in gas estimation from the behavior of geth:
+
+- Unlike geth, it is impossible to track out of gas errors on zkSync Era. The main reason is that the “actual” execution will happen inside the DefaultAccount system contract and due to the 63/64 rule when a high number of gas is provided, the call to the `execute` method of the DefaultAccount will NOT fail even though the subcall to the `transaction.to` contract did fail with an out of gas error.
```
StanislavBreadless, line 94:
```suggestion
- Unlike geth, it is impossible to track out of gas errors on zkSync Era. The main reason is that the “actual” execution will happen inside the DefaultAccount system contract and due to the 63/64 rule when a high number of gas is provided, the call to the `execute` method of the DefaultAccount will NOT fail due to out of gas even though the subcall to the `transaction.to` contract did fail with an out of gas error.
```

docs/dev/developer-guides/transactions/fee-model.md
```diff
@@ -39,3 +36,67 @@ So the tips to make the most out of the zkSync fee system are the following:
...
+
+That means that while geth can distinguish between “empty error” and “out of gas”, we can’t. If during gas estimation we fail to find the necessary `gasLimit`, will return the following error _“can not estimate gas: the transaction either fails or requires more gas than allowed”_.
```
StanislavBreadless, line 96: "the following error will be returned" or "the API will return ..."

docs/dev/developer-guides/transactions/fee-model.md
```diff
@@ -39,3 +36,49 @@ So the tips to make the most out of the zkSync fee system are the following:
...
+
+The `eth_estimateGas` method itself will use binary search to find the smallest gas value under which the transaction *succeeds*.
+
+### Difference from geth
+
+Just like geth, we will use binary search. However, there will be some notable differences in gas estimation from the behavior of geth:
+
+- Unlike geth, it is impossible to track out of gas errors on zkSync Era. The main reason is that the “actual” execution will happen inside the DefaultAccount system contract and due to the 63/64 rule when a high number of gas is provided, the call to the `execute` method of the DefaultAccount will NOT fail due to out of gas even though the subcall to the `transaction.to` contract did fail with an out of gas error.
+
+That means that while geth can distinguish between “empty error” and “out of gas”, we can’t. If during gas estimation we fail to find the necessary `gasLimit`, the following error will be returned _“can not estimate gas: the transaction either fails or requires more gas than allowed”_.
```
StanislavBreadless, line 78: I checked with the code, the error message is different from the one in the doc, so let's remove this sentence altogether to not confuse people

docs/dev/developer-guides/transactions/fee-model.md
```diff
@@ -39,3 +36,45 @@ So the tips to make the most out of the zkSync fee system are the following:
...
+## Gas estimation during transaction: For custom and paymaster accounts
+
+On Ethereum, there is a constant of 21000 gas that should cover all the intrinsic costs of processing a transaction: checking the signature and updating the nonce of the account. 
+
+On zkSync Era, it's quite different, because we support custom accounts that may take a slightly different (typically a bit higher) number of gas than EOAs. 
+We provide a convenient way for anyone to estimate the cost of a transaction regardless of the type of account. 
+
+### Changes in the `validateTransaction`
+
+At present, the `validateTransaction` method fails if the signature is incorrect and returns `success=true` for a valid signature. The system entirely disregards the `returndata`. However, we migrated to a new approach, where the success of the `validateTransaction` method depends on it returning a magic string. Any other `returndata` will be deemed invalid and rejected by the system.
```
uF4No, line 50: I think we should not mention how it was before (returning sucess=true) and just explain what's expected from the validateTransaction method now

StanislavBreadless, line 50: > At present

It is not how it is done presently, this is how it used to be. The sentence says that "we migrated to a new approach" :)

docs/.vuepress/config.ts
```diff
@@ -1,7 +1,7 @@
 import { defineUserConfig } from "vuepress";
 import docsearchPlugin from "@vuepress/plugin-docsearch";
 import theme from "./theme.js";
-import { pwaPlugin } from '@vuepress/plugin-pwa'
+//import { pwaPlugin } from '@vuepress/plugin-pwa'
```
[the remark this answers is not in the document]
krofax: It breaks on my local, I'll uncomment this right away.

docs/dev/developer-guides/transactions/fee-model.md
krofax: I rephrased to 

> The `validateTransaction` method fails if the signature is incorrect and returns `success=true` for a valid signature. The system entirely disregards the `returndata`. However, the success of the `validateTransaction` method depends on it returning a magic string. Any other `returndata` will be deemed invalid and rejected by the system.

@StanislavBreadless make sense?

docs/dev/developer-guides/transactions/fee-model.md
```diff
@@ -39,3 +36,45 @@ So the tips to make the most out of the zkSync fee system are the following:
 - **Reuse as many storage slots as possible.** Only the state diff is published on Ethereum.
 - **Users should share as many storage slots as possible.** If 100 users update a storage slot of your contract in a single block, the diff will be published only once. In the future, we will introduce reimbursement for the users, so that the costs for updating shared storage slots are split between the users.
 - **Reuse the contract code if possible.** On Ethereum, avoiding constructor parameters and putting them into constants reduces some of the gas costs upon contract deployment. On zkSync the opposite is true: deploying the same bytecode for contracts, while changing only constructor parameters can lead to substantial fee savings.
+
+
+## Gas estimation during transaction: For custom and paymaster accounts
```
githubdoramon, line 41: reads better if we swap "For Paymaster and Custom Accounts:

docs/dev/developer-guides/transactions/fee-model.md
```diff
@@ -39,3 +36,45 @@ So the tips to make the most out of the zkSync fee system are the following:
...
+
+### Notes on custom accounts
+
+Currently, the `validateTransaction` method of the AA (or the paymaster `validateAndPayForPaymasterTransaction` method) always tries to perform the same amount of computation (including storage accesses) regardless of whether the transaction is validated correctly. By default, the operator provides a transaction structure with the available information during fee estimation. To replace the signature, an invalid 65-byte ECDSA signature is utilized. The `DefaultAccount` (used by EOAs), during fee estimation, executes as many operations, including signature verification, and returns only `bytes4(0)` instead of magic. In the case of a custom account with multiple signers, the account may wish to simulate signature validation for all the provided signers.
+
+The code of the validation step of each account can be found in the [DefaultAccount implementation](https://github.com/matter-labs/era-system-contracts/blob/5a6c728576de5db68ad577a09f34e7b85c374192/contracts/DefaultAccount.sol#L65)
```
githubdoramon, line 56: Lets link to the file directly and not a specific version - https://github.com/matter-labs/era-system-contracts/blob/main/contracts/DefaultAccount.sol

docs/dev/developer-guides/transactions/fee-model.md
```diff
@@ -39,3 +36,45 @@ So the tips to make the most out of the zkSync fee system are the following:
...
+
+### Notes on the transaction’s length
+
+zkSync Era sends state diffs onchain, but the cost for the transaction will still mildly depend on its length (because long transactions need to be stored in the memory of the operator), also long transactions incur additional costs during interactions with an account. However, the signature (as well as its length) is not available at the time of fee estimation and so there is no correct way to precisely estimate the cost of the transaction. For now, we will compensate for it by multiplying the recommended cost of the transaction by a few percent. In the future, we may introduce the following:
```
githubdoramon, line 60: - memory of the operator), also long transactions incur additional costs during interactions with an account.
 
 I think we should have a period and not comma between these 2

docs/dev/developer-guides/transactions/fee-model.md
StanislavBreadless: @krofax 

How it used to be:

- `validateTransaction` is considered successful whenever it does not revert (i.e. returns `success = true`). For invalid signatures the method reverts.

How it is now:

- `validateTransaction` is considered successful whenever it does not revert (i.e. returns `success = true`) AND it returns the magic string. For invalid signatures, the method does not revert, but it returns invalid magic.

A lot of time has passed since the change, so we can only describe the part of how it is now.

Also, saying that the returndata is disregarded, but then saying that it is used to check the magic sounds a bit contradictive, but up to you

krofax: @StanislavBreadless I updated this part to be 

> The `validateTransaction` method is considered successful whenever it does not revert (i.e. returns `success = true`). And it returns the magic string. For invalid signatures, the method does not revert, but it returns invalid magic. However, the success of the `validateTransaction` method depends on it returning a magic string.

What do you think?

StanislavBreadless (approved):

docs/dev/building-on-zksync/hello-world.md
```diff
@@ -561,7 +578,7 @@ Even though ether is the only token you can pay fees with, the account abstracti
 
 ::: warning Paymasters on mainnet
 
-🚨 The testnet paymaster is purely for demonstration of this feature and won't be available on mainnet. When integrating your protocol on mainnet, you should follow the documentation of the paymaster you will use or create your own.
+The testnet paymaster is purely for demonstration of this feature and won't be available on mainnet. When integrating your protocol on mainnet, you should follow the documentation of the paymaster you will use or create your own.
```
AnastasiiaVashchuk, line 581: suggestion...I would highlighted  `The testnet paymaster` and make `paymaster`(used in this part "you should follow the documentation of the paymaster you will use or create your own.") as a link - [paymaster](../developer-guides/aa.md#paymasters)

docs/dev/building-on-zksync/hello-world.md
```diff
@@ -28,18 +28,22 @@ Please note that breaking changes were introduced in `zksync-web3 ^0.13.0`. The
...
 
-# For Yarn
 yarn init -y
 yarn add -D typescript ts-node ethers@^5.7.2 zksync-web3 hardhat @matterlabs/hardhat-zksync-solc @matterlabs/hardhat-zksync-deploy
 
-# For NPM
 npm init -y
 npm i -D typescript ts-node ethers@^5.7.2 zksync-web3 hardhat @matterlabs/hardhat-zksync-solc @matterlabs/hardhat-zksync-deploy
 ```
+:::
```
AnastasiiaVashchuk, line 46: image

AnastasiiaVashchuk: in `yarn` tab here is the instruction for npm too. Is has to be changed. npm tab shoud be added

AnastasiiaVashchuk: @krofax

AnastasiiaVashchuk: please do as you made it here
image

docs/api/hardhat/plugins.md
```diff
@@ -16,13 +16,22 @@ List of supported plugins:
...
+
+@tab npm
 
-# Npm (version 7 or later is recommended)
+```bash
 npm i -D @matterlabs/hardhat-zksync-toolbox
 ```
+:::
+
+> Note: For `npm`, version 7 or later is recommended.
```
AnastasiiaVashchuk, line 34: same issue as I mentioned [here](https://github.com/matter-labs/zksync-web-era-docs/pull/318/files#r1147336407) 
image

docs/dev/tutorials/custom-paymaster-tutorial.md
```diff
@@ -9,20 +9,34 @@ Let's see how we can use the paymaster feature to build a custom paymaster that
...
+@tab:active folder
+
+```sh
 mkdir custom-paymaster-tutorial
 cd custom-paymaster-tutorial
+```
+
+@tab:active yarn
+
+```bash
```
AnastasiiaVashchuk, line 35: on this page, below still are some commands that are without `@tab:active yarn`

docs/dev/tutorials/custom-aa-tutorial.md
```diff
@@ -5,23 +5,36 @@ In this tutorial, we build a factory that deploys 2-of-2 multisig accounts.
...
+::: code-tabs
+
+@tab:active folder
+
+```sh
 mkdir custom-aa-tutorial
 cd custom-aa-tutorial
+```
+
+@tab yarn
```
AnastasiiaVashchuk, line 31: on this page, below still are some commands that are without @tab:active yarn

docs/dev/tutorials/cross-chain-tutorial.md
```diff
@@ -155,14 +155,20 @@ main().catch((error) => {
 
 5. Compile the contract and run the deployment script with:
 
-```
-# compile contract
+::: code-tabs
+
+@tab:active Compile contract
```
AnastasiiaVashchuk, line 160: please check the whole page and refactor all snippets with yarn, sh/bash etc
for example here yarn commands are still in text, is  some other places they are already in *yarn* tab active

AnastasiiaVashchuk (commented):
```
 ::: warning
This section of the documentation is under review to reflect the changes made to the system contracts ([see changelog](../../troubleshooting/changelog.md)). A revised version will be available shortly.
:::

```

is this warning still  actual?

docs/dev/developer-guides/transactions/fee-model.md
```diff
@@ -39,3 +36,45 @@ So the tips to make the most out of the zkSync fee system are the following:
...
+
+- Unlike geth, it is impossible to track out of gas errors on zkSync Era. The main reason is that the “actual” execution will happen inside the DefaultAccount system contract and due to the 63/64 rule when a high number of gas is provided, the call to the `execute` method of the DefaultAccount will NOT fail due to out of gas even though the subcall to the `transaction.to` contract did fail with an out of gas error.
+
+- During simulation, geth uses `tx.gasprice = 0` to make sure that the user can pay the fee even though the `tx.origin` in the simulation may not have any balance at all - This means that when `estimateGas` from an empty account is called, no `value` can be provided to such call as this account has zero balance to cover this value. 
+
+We could do that, but that would mean that the `payForTransaction` of the Account Abstraction protocol would do nothing and thus be much cheaper than it will be during the actual transaction validation. Instead, the operator will increase the balance of the user’s account by `tx.maxFeePerErg * tx.ergsLimit`.
```
AnastasiiaVashchuk, line 78:
```suggestion
We could do that, but that would mean that the `payForTransaction` of the Account Abstraction protocol would do nothing and thus be much cheaper than it will be during the actual transaction validation. Instead, the operator will increase the balance of the user’s account by `tx.maxFeePerGas * tx.gasLimit`.
```

AnastasiiaVashchuk, line 78: is this related to second point "During simulation, geth ...."? 
then it should be in the second section.   
Now it looks  to be separate 
image

AnastasiiaVashchuk (commented):
>zkSync's version of gas is called gas just like Ethereum and represents not only the costs of computations but also the cost of publishing data on-chain and affecting storage.


need to be changed :

`zkSync's version of gas represents not only the costs of computations but also the cost of publishing data on-chain and affecting storage.`

AnastasiiaVashchuk (commented):
>the number of gas needed for


need to be changed :

`the  amount of gas needed for`

docs/dev/developer-guides/transactions/fee-model.md
```diff
@@ -39,3 +36,45 @@ So the tips to make the most out of the zkSync fee system are the following:
...
+
+### Changes for `eth_estimateGas`
+
+Since the entire flow for tx validation + execution will be executed to get the transaction’s fee, the operator will manually put the necessary balance (if needed) for the account before the emulation is run to conduct the default account’s `payForTransaction` method.
+
+The `eth_estimateGas` method itself will use binary search to find the smallest gas value under which the transaction *succeeds*.
+
+### Difference from geth
+
+Just like geth, we will use binary search. However, there will be some notable differences in gas estimation from the behavior of geth:
```
AnastasiiaVashchuk, line 72:
```suggestion
Just like Geth, we will use binary search for gas estimation. However, there will be some notable differences in gas estimation from the behaviour of Geth:
```

docs/dev/developer-guides/transactions/fee-model.md
```diff
@@ -39,3 +36,45 @@ So the tips to make the most out of the zkSync fee system are the following:
...
+
+- Unlike geth, it is impossible to track out of gas errors on zkSync Era. The main reason is that the “actual” execution will happen inside the DefaultAccount system contract and due to the 63/64 rule when a high number of gas is provided, the call to the `execute` method of the DefaultAccount will NOT fail due to out of gas even though the subcall to the `transaction.to` contract did fail with an out of gas error.
```
AnastasiiaVashchuk, line 74:
```suggestion
- Unlike Geth, it is impossible to track  *out of gas* errors on zkSync Era. The main reason is that the “actual” execution will happen inside the DefaultAccount system contract and due to the 63/64 rule when a high number of gas is provided, the call to the `execute` method of the DefaultAccount will NOT fail due to out of gas even though the subcall to the `transaction.to` contract did fail with an out of gas error.
```

docs/dev/developer-guides/transactions/fee-model.md
```diff
@@ -39,3 +36,45 @@ So the tips to make the most out of the zkSync fee system are the following:
...
+
+- During simulation, geth uses `tx.gasprice = 0` to make sure that the user can pay the fee even though the `tx.origin` in the simulation may not have any balance at all - This means that when `estimateGas` from an empty account is called, no `value` can be provided to such call as this account has zero balance to cover this value.
```
AnastasiiaVashchuk, line 76:
```suggestion
- During simulation, Geth uses `tx.gasprice = 0` to make sure that the user can pay the fee even though the `tx.origin` in the simulation may not have any balance at all. This means that when `estimateGas` from an empty account is called, no `value` can be provided to such call as this account has zero balance to cover this value. 
```

docs/dev/developer-guides/transactions/fee-model.md
AnastasiiaVashchuk: for me

PR #331, opened 2023-03-16
Remove version from zksync-web3 which allows for latest version

cooganb:
Removes version from zksync-web3 to prevent some issues we've seen in Discord at least once of someone locking to an earlier version.


cooganb:
@krofax @uF4No

krofax (approved):

uF4No:
we merged the same change in a different PR

PR #346, opened 2023-03-22
Small fixes to hello-world paymaster code example

pauldowman:
- Remove duplicated comment (left it before the whole block, instead of in both places) 
- Fix formatting (indentation)
- Add `toString()` to `gasLimit` in both examples (it's necessary, otherwise it fails with a Paymaster validation error when submitting the tx)


kailash-manasarovar:
Hi @pauldowman apologies for the delay on this one. It looks like we made your suggested change(s) in an earlier update. Thanks so much for your input.

PR #352, opened 2023-03-23
Update system-contracts.md - 

doganalpaslan:
Since you have open sourced the bootloader contract it needs to be updated, some devs or researchers who wants to build on a top of zkSync Era may be confused with this paragraph: "For now, you do not have to know any details about it, but knowing that it exists is important when you develop using the account abstraction feature. You can always assume that the bootloader is not malicious and it is a part of the protocol. In the future, the code of the bootloader will be made public and any changes to it will also mean an upgrade to the protocol." 
It would be better to share the audit and the codebase at your docs. 
[#](https://era.zksync.io/docs/dev/developer-guides/system-contracts.html#protected-access-to-some-of-the-system-contracts)

### DIFF
--- a/docs/dev/developer-guides/system-contracts.md
+++ b/docs/dev/developer-guides/system-contracts.md
@@ -34,9 +34,9 @@ This contract stores account nonces. The account nonces are stored in a single p
 
 ## Bootloader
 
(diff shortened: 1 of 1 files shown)


CoinKafasi:
I totally agree with your friend.

kailash-manasarovar (approved):

PR #355, opened 2023-03-24
Fix small typo in interacting.md

mmqxyz:


uF4No:
@kailash-manasarovar can you remove this phrase now that we're fully open to everyone?

PR #359, opened 2023-03-25
change go-sdk: go-ethereum/rpc to common

egeaybars123:
I was going through the go-sdk docs for zkSync Era, and found out that the import parts of the go scripts could need a change. I saw that some scripts were importing go-ethereum/rpc although it was supposed to import go-ethereum/common. In the general script, rpc is needed, but since the docs were organized according to single functions (deposit, withdraw etc.), not importing the necesssary packages could confuse developers. 

I changed the go-ethereum/rpc imports to go-ethereum/common wherever they were supposed to be imported.

### DIFF
--- a/docs/api/go/getting-started.md
+++ b/docs/api/go/getting-started.md
@@ -76,7 +76,7 @@ ethereumProvider, err := w.CreateEthereumProvider(ethRpc)
 package main
 
 import (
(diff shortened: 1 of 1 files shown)


githubdoramon (approved):

githubdoramon:
Thanks for contributing @egeaybars123

PR #366, opened 2023-03-28
update gitignore

kailash-manasarovar:
- Add ignore to fix unwanted upgrades in pushes.
- The `DS_Store` remove is a duplicate.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,10 +41,8 @@ lib
 /report.txt
 /tokens.txt
 
-!.yarn/releases/*
-!.yarn/plugins/*
+.yarn/releases/*
+.yarn/plugins/*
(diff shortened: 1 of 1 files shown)


githubdoramon (changes requested):
There are 2 .temp, and would keep the .DS_Store because every Mac creates this file

kailash-manasarovar:
> There are 2 .temp, and would keep the .DS_Store because every Mac creates this file

- There were 2 DS_Stores, I removed one. Scroll up and you'll see the other one.
- .temp vs .temp/ appears to be protecting against a temp **file** and a temp **directory**.

githubdoramon (approved):

PR #349, opened 2023-03-23
Cross chain governance tutorial - clarifying and fixing

kailash-manasarovar:
This is working and clarified up to **Calling the contract L1->L2**, which is erroring and with devs who are taking a look. So 95% 

* Tutorial has both npm/yarn instructions.
* For each option, the resulting error at **Calling the contract L1->L2** is the same `BigNumber` error.

PR #371, opened 2023-03-31
fix: cross chain tutorial

samlior:
It looks like there is an argument missing when calling `callZkSync`.


samlior:
By the way, if `gasLimit` is set to 100000, the transaction will be rolled back on [this line](https://github.com/matter-labs/v2-testnet-contracts/blob/40cc06e6f3c54232c666590dc71c6cf68882299b/l1/contracts/zksync/facets/Mailbox.sol#L347). After I changed `gasLimit` to 300000, the transaction was successfully packaged on L1, but still failed on L2, even if I set `gasLimit` to 400000 doesn't work either. I don't know why.

Also I am curious, if the transaction fails, how should I get back the amount I sent and the remaining handling fee?

samlior:
@kailash-manasarovar #349 This is not a `BigNumber` error, because `ethers` tries to parse the last input parameter into a `BigNumber` because one parameter is missing when calling, but in fact the last input parameter is `value` and `gasPrice`, so When parsing you get an error.

kailash-manasarovar:
> @kailash-manasarovar #349 This is not a `BigNumber` error, because `ethers` tries to parse the last input parameter into a `BigNumber` because one parameter is missing when calling, but in fact the last input parameter is `value` and `gasPrice`, so When parsing you get an error.

Thank you @samlior for the input :) Our devs are looking at this script at the moment. I am going to run your suggested update now. Please check again in a bit for updates.

kailash-manasarovar (approved):

kailash-manasarovar:
> By the way, if `gasLimit` is set to 100000, the transaction will be rolled back on [this line](https://github.com/matter-labs/v2-testnet-contracts/blob/40cc06e6f3c54232c666590dc71c6cf68882299b/l1/contracts/zksync/facets/Mailbox.sol#L347). After I changed `gasLimit` to 300000, the transaction was successfully packaged on L1, but still failed on L2, even if I set `gasLimit` to 400000 doesn't work either. I don't know why.
> 
> Also I am curious, if the transaction fails, how should I get back the amount I sent and the remaining handling fee?

kailash-manasarovar:
> By the way, if `gasLimit` is set to 100000, the transaction will be rolled back on [this line](https://github.com/matter-labs/v2-testnet-contracts/blob/40cc06e6f3c54232c666590dc71c6cf68882299b/l1/contracts/zksync/facets/Mailbox.sol#L347). After I changed `gasLimit` to 300000, the transaction was successfully packaged on L1, but still failed on L2, even if I set `gasLimit` to 400000 doesn't work either. I don't know why.
> 
> Also I am curious, if the transaction fails, how should I get back the amount I sent and the remaining handling fee?

Will get back to you on this question.

PR #373, opened 2023-03-31
fix returned tokens format

uF4No:
Quick fix reported by a user.

### DIFF
--- a/docs/api/api.md
+++ b/docs/api/api.md
@@ -106,7 +106,7 @@ None.
 
 ### `zks_getConfirmedTokens`
 
-Given `from` and `limit` return information about the confirmed tokens with IDs in the interval `[from..from+limit-1]`. "Confirmed" is a misnomer here, since a confirmed token is one that has been bridged through the default zkSync Era bridge. This method will mostly be used by the zkSync team internally.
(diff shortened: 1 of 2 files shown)


docs/api/api.md
```diff
@@ -106,7 +106,7 @@ None.
 
 ### `zks_getConfirmedTokens`
 
-Given `from` and `limit` return information about the confirmed tokens with IDs in the interval `[from..from+limit-1]`. "Confirmed" is the wrong word here, since a confirmed token has already been bridged through the default zkSync Era bridge.
+Given `from` and `limit` return information about the confirmed tokens with IDs in the interval `[from..from+limit-1]`. "Confirmed" is a misnomer here, since a confirmed token is one that has been bridged through the default zkSync Era bridge. This method will mostly be used by the zkSync team internally.
```
kailash-manasarovar, line 109:
```suggestion
`from` and `limit` return information about the confirmed tokens with IDs in the interval `[from..from+limit-1]`. **Confirmed** is a misnomer here, since a confirmed token is one that has been bridged through the default zkSync Era bridge. This method will mostly be used by the zkSync team internally.
```

kailash-manasarovar: You have a given clause without a resolution. Better to remove it.

kailash-manasarovar (approved):

PR #382, opened 2023-04-05
Cross-chain tutorial: update compiler and test again

kailash-manasarovar:
* Updated the compiler to latest release.
* Tested the flow again, all good.
* Minor tweaks to language.

### DIFF
--- a/docs/dev/tutorials/cross-chain-tutorial.md
+++ b/docs/dev/tutorials/cross-chain-tutorial.md
@@ -10,9 +10,10 @@ This tutorial shows you how to implement communication between L1 and L2 with th
 - You are already familiar with deploying smart contracts on zkSync. If not, please refer to the first section of the [quickstart tutorial](../building-on-zksync/hello-world.md).
 - You already have some experience working with Ethereum.
 - You have a web3 wallet app which holds some Goerli test ETH and some zkSync test ETH.
+- You know how to get your [private key from your MetaMask wallet](https://support.metamask.io/hc/en-us/articles/360015289632-How-to-export-an-account-s-private-key).
 
-:::warning
-- The `yarn` instructions run with Node version 14.
+:::tip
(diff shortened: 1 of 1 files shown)


krofax (dismissed):

uF4No:
Steps 2 & 3 of the L1 part are not neccessary as `npx hardhat` does the `npm init /yarn init` and being and npx script, there's no need to install hardhat beforehand.

Also, step 3 in the L2 part is not necessary. Using`npx hardhat` for a zkSync-specific project is not recommended and will mess-up the dependencies. It's better to just initialise a project with `npm init/yarn init` and install the dependencies manually (step 2). Then create the hardhat.config.ts manually and the contracts. Avoid using `npx hardhat`.

Remove the warning for "Error: Bytecode length in 32-byte words must be odd,"

kailash-manasarovar:
> Steps 2 & 3 of the L1 part are not neccessary as `npx hardhat` does the `npm init /yarn init` and being and there's no need to install hardhat beforehand.

`npx hardhat` does an `npm` default install, not yarn. I am referring to Hardhat docs: https://hardhat.org/hardhat-runner/docs/guides/project-setup.
> 
> Also, step 3 in the L2 part is not necessary. Using`npx hardhat` for a zkSync-specific project is not recommended and will mess-up the dependencies. It's better to just initialise a project with `npm init/yarn init` and install the dependencies manually (step 2). Then create the hardhat.config.ts manually and the contracts. Avoid using `npx hardhat`.
> 
I added it in because it sets up files and folders for you, which makes life easier. Also, there is a bit of a disconnect between part 1 and part 2 if you don't do it in part 2. It works too. We can reconvene on it.

> Remove the warning for "Error: Bytecode length in 32-byte words must be odd,"

Will do.

uF4No (approved):

PR #383, opened 2023-04-05
New PWA config to handle updates

githubdoramon:


docs/.vuepress/config.ts
```diff
@@ -1,7 +1,7 @@
 import { defineUserConfig } from "vuepress";
 import docsearchPlugin from "@vuepress/plugin-docsearch";
 import theme from "./theme.js";
-import { pwaPlugin } from '@vuepress/plugin-pwa'
+import { pwaPlugin } from "vuepress-plugin-pwa2";
```
uF4No, line 4: woudnt this fail? as the dependency in package.json is still "@vuepress/plugin-pwa"

githubdoramon: The dependency was removed from package.json (and this one comes bundled in the hope theme)

krofax (approved):

uF4No (approved):

PR #423, opened 2023-04-14
Readme: Add quotes around `ethers` yarn install command

makevoid:
Added `"` quotes to fix install error in zsh

sample output of the error:
```
(base) makevoid@box zksync_project % yarn add -D ethers@^5.7.2
zsh: no matches found: ethers@^5.7.2
```

### DIFF
--- a/docs/dev/building-on-zksync/hello-world.md
+++ b/docs/dev/building-on-zksync/hello-world.md
@@ -33,7 +33,7 @@ cd greeter-example
 
 ```sh
 yarn init -y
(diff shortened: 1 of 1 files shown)


uF4No:
we havent been able to replicate the issue with Yarn v1 or v3. Which version are you using?

makevoid:
I found that this is because I have zsh with `extended_glob` enabled. Disabling it fixed the issue. 
I think there's no harm in adding quotes so that the users that have this setting enabled don't get this error.

more info on zsh extended glob here:
https://zsh.sourceforge.io/Doc/Release/Options.html -> ctrl + f - EXTENDED_GLOB


btw the quickstart docs are great, thanks for creating them

kailash-manasarovar:
Thank you.

PR #430, opened 2023-04-17
Trigger Algolia reindex after prod deployment

hatemosphere:


kailash-manasarovar (approved):
Thanks so much for this @hatemosphere :)

PR #444, opened 2023-04-21
Updated/tested tutorial compatibility with compiler version.

krofax:
Updated and tested tutorials(except aa daily limit and paymasters) compatibility with the compiler `v1.3.9`.


kailash-manasarovar (approved):
These are OK.

PR #453, opened 2023-04-24
fixes to match interfaces methods

uF4No:
Fix to match methods in the docs with the methods defined in the [IAccount](https://github.com/matter-labs/v2-testnet-contracts/blob/main/l2/system-contracts/interfaces/IAccount.sol) and [IPaymaster](https://github.com/matter-labs/v2-testnet-contracts/blob/main/l2/system-contracts/interfaces/IPaymaster.sol) interfaces


kailash-manasarovar (approved):
LGTM

PR #455, opened 2023-04-25
Adds metadata settings field to zksolc config

mpopovac-txfusion:


uF4No:
To be merged with new release of hardhat-zksync-solc

PR #462, opened 2023-04-26
fix python sdk TxFunctionCall import examples (#461)

### LINKED ISSUE #461: Typo in python examples (TxFunctionCall import)
## Overview
At multiple points in the python sdk docs, the `TxFunctionCall` object is imported from `zksync2.transaction.transaction712` but it does not exist at this location.

## Solution
All instances of:
```py
from zksync2.transaction.transaction712 import TxFunctionCall
```
should be replaced with:
```py
from zksync2.transaction.transaction_builders import TxFunctionCall
```

omurovec:
fixes #461


kailash-manasarovar:
Hi @omurovec, we're about to update the whole Python SDK, including docs, so this change may be already in the pipeline. I'm going to close your PR for now but thank you for raising :)

PR #464, opened 2023-04-27
Add warning to compiler toolchain

krofax:
Add warning to compiler toolchain page.


kailash-manasarovar (changes requested):
Why only on this page?

Add here: https://era.zksync.io/docs/api/compiler-toolchain/
And here: https://era.zksync.io/docs/api/compiler-toolchain/vyper.html

And check with @hedgar2017 if you should add here too.

hedgar2017 (dismissed):

hedgar2017:
> Why only on this page?
> 
> 
> 
> Add here: https://era.zksync.io/docs/api/compiler-toolchain/
> 
> And here: https://era.zksync.io/docs/api/compiler-toolchain/vyper.html
> 
> 
> 
> And check with @hedgar2017 if you should add to LLVM.

LLVM is too much because it may be unclear what version it refers too. Overview, zksolc and zkvyper must be enough.

hedgar2017 (approved):

PR #466, opened 2023-04-28
fix: modify import path in demo code of python api docs.

a3roger:
- The project [zksync2-python](https://github.com/zksync-sdk/zksync2-python) has update to `V0.4.0`, but the docs is not updated.
- This commit updates two places: 
  1. modify import path in demo code of python api docs
  2. modify `Web3.toWei` to `Web3.to_wei`


kailash-manasarovar:
Hi @a3roger we are in the process of completely updating the Python SDK. Please keep your eye on the docs for the update https://era.zksync.io/docs/api/python. Thanks for the PR.

PR #484, opened 2023-05-05
Added footer to docs

krofax:
Added footer to docs.


kailash-manasarovar:
This only works on the front page. All other pages aren't working.

kailash-manasarovar:
Also @krofax you are the only contributor on every page again here, so something from your other PR crept in.

kailash-manasarovar:
> Also @krofax you are the only contributor on every page again here, so something from your other PR crept in.

Apologies, I'm wrong about this. I think this is default behavior with previews.

kailash-manasarovar (approved):
@krofax 

I have updated a few things:

1. I removed the script which was not the way to deal with the bleeding toc. 
2. I fixed the footer links.
3. I removed the contact page.
4. I fixed some minor things.

I'm approving because it is 1M x better than what we have now. The continued bleeding toc we can live with while we get an expert to look at the problem. It is only on a few pages. Manually adding a height value to each page was not the answer. I expect the issue is with a missing <div> on the page which an expert web dev will see immediately.

krofax:
> @krofax
> 
> I have updated a few things:
> 
> 1. I removed the script which was not the way to deal with the bleeding toc.
> 2. I fixed the footer links.
> 3. I removed the contact page.
> 4. I fixed some minor things.
> 
> I'm approving because it is 1M x better than what we have now. The continued bleeding toc we can live with while we get an expert to look at the problem. It is only on a few pages. Manually adding a height value to each page was not the answer. I expect the issue is with a missing `<div>` tag on the page which an expert web dev will see immediately.
> 
> Let me know if you have any questions. Have a look then go ahead and merge 😃

Totally agree with you on it.
It took me some time and still wasn't able to get it fixed.

PR #514, opened 2023-05-17
Add social share preview link to docs

krofax:
Added social share preview link to docs.
Unfortunately, we can't preview on staging, only on prod.


obmatter (approved):
Reviewed from my side. 
Looks ok. Can be checked only after release to prod. 
Please, ping me to check.

PR #516, opened 2023-05-17
Update redirection url

krofax:
Updated redirection url to footer anchor.

Test with https://aqwzx-zksync-v2-docs--pr516-fix-contact-links-do-f6lppu79.web.app/docs/contact.html


kailash-manasarovar:
I don't think this is the way @krofax . An anchor link has to be an <a href="#my-anchor-link"> text </a>. You are trying to make a css id an anchor link and this won't work. You will have to add an anchor link to the footer.

kailash-manasarovar (approved):
Yes, seems to be a good option.

PR #541, opened 2023-05-30
Update bridging-asset.md

0x3c:
Fix wrong method name, see [IL1Bridge.sol -> claimFailedDeposit](https://github.com/matter-labs/era-contracts/blob/main/ethereum/contracts/bridge/interfaces/IL1Bridge.sol#L40)


kailash-manasarovar (approved):
LGTM. Thank you :)

PR #567, opened 2023-06-03
Add private key best practices as a prerequisite

damien75:
We should include some documentation to make sure people follow best practices when using private keys to deploy their smart contracts. Optional could be to suggest to use a new account that will be utilized for testing purposes only


kailash-manasarovar:
Thanks. This is a good idea. However, I would rather add a How To for dotenv and link out to it from all our tutorials/quick starts/etc... watch this space. Thanks again.

PR #608, opened 2023-06-24
fix: add documentation for account abstraction

danijelTxFusion:


uF4No:
Can we add some internal links so the code examples in the getting started are easier to find? For example:

- In the accounts page, add links to the examples to do deposits and withdraws.
- In the paymaster page, add links to the paymaster examples.

danijelTxFusion:
> Can we add some internal links so the code examples in the getting started are easier to find? For example:
> 
>     * In the accounts page, add links to the examples to do deposits and withdraws.
> 
>     * In the paymaster page, add links to the paymaster examples.

I added links to those pages.

PR #615, opened 2023-06-30
Proxy fee estimation feature, verify plugin core dependency change

FilipTxFusion:
- Add proxy fee estimation feature for the upgradable plugin
- Change verify plugin core dependency to @nomicfoundation/hardhat-verify


docs/tools/hardhat/hardhat-zksync-upgradable.md
```diff
@@ -646,4 +646,65 @@ This command will verify the implementation related to the proxy, the proxy cont
...
+
+If you want to get the estimation for the beacon proxy contract, please use the `estimateGasBeaconProxy` method:
+
+``` typescript
+const totalGasEstimation = await hre.zkUpgrades.estimation.estimateGasBeacon(deployer, contract, []);
+```
+
+
+All of those methods calculate the fee for every contract in their corresponding pipeline, print that cost on the console and return the total sum.
+If you don't want to see all the separate estimations, you can use the parameter `quiet` as the last parameter in every method and receive just the returned sum.
```
dutterbutter, line 706:
```suggestion
Each of these methods totals the fee for every contract in their respective pipeline, displays the cost on the console, and returns the cumulative sum. If you prefer not to see the individual estimations, introduce the parameter `quiet` as the final parameter in any method to receive only the returned sum.
```

dutterbutter (changes requested):
Few small changes

FilipTxFusion:
> Few small changes

Agree with all of them. Committed the changes.

dutterbutter:
@FilipTxFusion please resolve conflicts and we should be good! Thanks 🙏

uF4No (approved):

dutterbutter (approved):

PR #625, opened 2023-07-05
SEO improvements

uF4No:
- Fixes broken menu in Go SDK
- Changes `aa.md` to `account-abstraction.md` for better SEO. Changes links and menus accordingly
- Fixes all redirects pointing to old paths
- Add some internal links
- Reorg the order of some menu items


docs/.vuepress/sidebar/en.ts
```diff
@@ -29,7 +30,7 @@ export const enSidebar = sidebar({
     },
     {
       text: "Tutorials",
-      link: "/dev/tutorials/cross-chain-tutorial.md",
+      link: "/dev/tutorials/README.md",
```
dutterbutter, line 33: Can you do the same for the "How to". That is, add a README with summary.

docs/api/js/utils.md
```diff
@@ -553,6 +554,34 @@ export function getHashedL2ToL1Msg(sender: Address, msg: BytesLike, txNumberInBl
 }
 ```
 
+
+### `getPaymasterParams`
```
dutterbutter, line 558: We have this same function documented in the [Paymaster Utilities](https://era.zksync.io/docs/api/js/paymaster-utils.html#getpaymasterparams) section. Is that intentional? If so, we should update that documented Paymaster Utilities to show full implementation as this one does.

dutterbutter (changes requested):
Small changes. Also not sure if you ran `yarn lint:dead` locally but some messaging indicating `File is not found` and `Hash should slugify` are presented.

docs/api/js/utils.md
uF4No: missed that, I'll revert this change

uF4No: fixed in https://github.com/matter-labs/zksync-web-era-docs/pull/625/commits/e9a625b0b820ad302d3b0eb7e78d2fa7e6970cd1

docs/.vuepress/sidebar/en.ts
uF4No: fixed in https://github.com/matter-labs/zksync-web-era-docs/pull/625/commits/be942f012e5c9caa23056d437a54f1babb8818c0

dutterbutter (approved):

PR #667, opened 2023-08-17
Update differences-with-ethereum.md

miladpiri:
There were some issues with this example:
 - All the immutables should be initialized. So, here we will get compile time error regarding `y` which is not initialized.
  - Using return in assembly, will left the values of immutables equal to their default value (in this case `0`). For example, in the mode Y+M3B3, the value `x` will be equal to `0` not `45`. For example, if the function `getData()` returns `x`, it will return `0`.
   - Using some other modes (like E-M0B2 or Y+M1B0), will produce the following compile-time error. To solve it, the immutable `x` should not be public. ``` Errors found: ["CodeGenerationError: Some immutables were read from but never assigned, possibly because of optimization. ```


hedgar2017 (approved):
LGTM.

idea404 (approved):

PR #673, opened 2023-08-18
Update tutorial links to monorepo links

idea404:
# What :computer: 
* This PR updates the links in the relevant parts of the docs that would link to the old tutorials repos. Now they point to the monorepo. 

# Why :hand:
* Monorepo for tutorials was created here: https://github.com/matter-labs/tutorials


MexicanAce (approved):
LGTM

PR #704, opened 2023-09-15
qa: “Send an L2 to L1 message” tutorial PR fix

obmatter:
[SUGGESTION]:💡 It would be great to add how users can import those files. 

image

It’s difficult to understand where should we get example files from. 

We don’t have clear steps for all those 1-4 steps - let’s specify what commands have to be run so as not to confuse users.

Same unclear steps for

image


MexicanAce:
Should this be a GitHub Issue instead of a PR?

PR #711, opened 2023-09-18
Update Docs to Include New Definition of Block Timestamp and Number

idea404:
# What :computer: 
* [Current protocol update](https://github.com/zkSync-Community-Hub/zkync-developers/discussions/87) will change definitions of both

# Why :hand:
* The old definitions will be wrong


docs/dev/tutorials/aa-daily-spend-limit.md
```diff
@@ -1040,15 +1040,7 @@ Reset time was not updated as not enough time has passed
...
-Since `ONE_DAY` is set to 1 minute for this test in the `SpendLimit.sol` contract, you should expect it to reset after 60 seconds. However, we're using `block.timestamp` so the limit is only reset after a new L1 batch is sealed (around ten minutes on testnet). To understand the reason behind this, we should know about the constraints around using `block.timestamp`.
-
-::: warning `block.timestamp` returns L1 batch value
-
-`block.timestamp` returns the time of the latest L1 batch, instead of the L2 block, and it's only updated once a new batch is sealed (~5-10 minutes on testnet). What this means is that `block.timestamp` in smart contracts on zkSync Era is currently a delayed value.
-
-To keep this tutorial as simple as possible, we've used `block.timestamp` but we don't recommend relying on this for accurate time calculations.
-
-:::
+Since `ONE_DAY` is set to 1 minute for this test in the `SpendLimit.sol` contract, you should expect it to reset after 60 seconds. However, we're using `block.timestamp` so the limit is only reset after a new L1 block is collated (around 2-3 seconds).
```
MexicanAce, line 1043: Should this say `L2 block`?
```suggestion
Since `ONE_DAY` is set to 1 minute for this test in the `SpendLimit.sol` contract, you should expect it to reset after 60 seconds. However, we're using `block.timestamp` so the limit is only reset after a new L2 block is collated (around 2-3 seconds).
```

MexicanAce (approved):
LGTM

PR #716, opened 2023-09-22
fix: redstone tutorial missing links

idea404:
# What :computer: 
* Fixes cd command to go one level lower
* Adds link to tutorial repo in text 

# Why :hand:
* cd was leading to higher level readme only


AlbionaHoti (approved):
LGTM

PR #730, opened 2023-09-29
feat: add JS Ethers V6 SDK

danijelTxFusion:
# What :computer: 
* Documentation for JavaScript Ethers V6 SDK has been added.


docs/.vuepress/sidebar/en.ts
```diff
@@ -114,6 +114,34 @@ export const enSidebar = sidebar({
         "/api/js/front-end",
       ],
     },
+    {
+      text: "JavaScript Ethers V6 SDK",
```
idea404, line 118: I find it would make it easier to follow if both the v5 and v6 docs are under the same "JS SDK" category on the sidebar, then we have a "JS SDK ev5" subcategory under it with all the docs from before and another "JS SDK ev6" category with these new docs, as:

JS SDK 
--ethersv5 
----old docs as before
--ethersv6
----these new docs

idea404 (changes requested):
One format adjustment and one spelling

docs/api/js/ethers-v6/getting-started.md
```diff
@@ -0,0 +1,92 @@
...
+import { ethers } from "ethers";
+
+const provider = Provider.getDefaultProvider(types.Network.Goerli);
+const ethProvider = ethers.getDefaultProvider("goerli");
+```
+
+Get the network:
+
+```ts
+console.log(`Network: ${await provider.getNetwork()}`);
```
elek, line 62: What is the expected output here?

After I added `types` to the import, I was able to execute this script with `npx ts-node src/hello.ts`, but the output is the following (which is not very useful). 

```
Network: [object Object]
```

docs/api/js/ethers-v6/getting-started.md
```diff
@@ -0,0 +1,92 @@
...
+  transactions.
+- `Wallet` wraps all operations that interact with an account. An account generally has a private key, which can be used to sign a variety of
+  types of payloads. It provides easy usage of the most common features.
+
+## Examples
+
+Connect to the zkSync Era network:
+
+```ts
+import { Provider, utils } from "zksync2-js";
```
elek, line 52: As a user, who is not native in the js/ts word, I would appreciate some explanation how these examples can be executed.

For example installing ts-node + `npx ts-node src/hello.ts`

docs/api/js/ethers-v6/examples/deposit.md
```diff
@@ -0,0 +1,87 @@
...
+const PRIVATE_KEY = process.env.PRIVATE_KEY;
+const wallet = new Wallet(PRIVATE_KEY, provider, ethProvider);
+
+async function main() {
+  // Crown token which can be mint for free
+  const tokenL1 = "0xc8F8cE6491227a6a2Ab92e67a64011a4Eba1C6CF";
+  const tokenL2 = await provider.l2TokenAddress(tokenL1);
+
+  console.log(`L2 balance before deposit: ${await wallet.getBalance(tokenL2)}`);
+  console.log(`L1 balance before deposit: ${await wallet.getBalanceL1(tokenL1)}`);
```
elek, line 66: I tested code, and It failed for me, potentially because an overflow error:

```
L2 balance before deposit: 0
L1 balance before deposit: 99999999999999999999999999990
Error: TypeError: underflow (argument="value", value=5e-13, code=INVALID_ARGUMENT, version=6.7.1)
```

My balance was 99999999999999999999999999990 with a test token where the decimal was 18 (99999999999.99999999999999999)

docs/api/js/ethers-v6/getting-started.md
elek: + I think the `types` also should be imported here.

docs/api/js/ethers-v6/examples/deposit.md
danijelTxFusion: Can you please provide your account address, token address so I can reproduce the error that you mentioned.

elek: Contract: 0xb24f060AEdEF6f64Abaaf33Dc72Bc699400681ea
Account: 0xAfD2d8d18B2B7D1C1f35869Df4ab66BAeFc9A912

Contract was something like this (but with more zero, I removed a few of them to fix the issue):

```
pragma solidity ^0.8.7;

import "node_modules/@openzeppelin/contracts/token/ERC20/ERC20.sol";

contract GLDToken is ERC20 {
    constructor() ERC20("Gold", "GLD") public {
        _mint(msg.sender, 1000000000000000000000);
    }
}
```

docs/.vuepress/sidebar/en.ts
danijelTxFusion: Suggested changes are applied.

docs/api/js/ethers-v6/getting-started.md
danijelTxFusion: I added `toJSON` helper function which prints any object: 

Get the network (helper function [toJSON](./providers.md#tojson)):

```ts
console.log(`Network: ${toJSON(await provider.getNetwork())}`);
```

docs/api/js/ethers-v6/getting-started.md
danijelTxFusion: `types` are imported. I also added following section:

## Examples

The following examples could be easily run by writing the code snippets in a file and executing them as follows:

```shell
npx ts-node src/zksync.ts
```

idea404 (approved):

docs/api/js/ethers-v6/examples/deposit.md
danijelTxFusion:
```
Error: TypeError: underflow (argument="value", value=5e-13, code=INVALID_ARGUMENT, version=6.7.1)
```
The error is not related to the smart contract; it is related to the value itself because it's passed as a floating-point number (`0.0000000000005`) instead of an integer.

PR #761, opened 2023-10-24
 chore: Update 'utilities library definition' link in utils.md

0xChristia:
# What :computer: 
- Update 'utilities library definition' link in utils.md

# Why :hand:
image

# Evidence :camera:
 No


idea404:
https://github.com/matter-labs/zksync-era/blob/main/sdk/zksync-web3.js/src/utils.ts link returns 404

PR #767, opened 2023-10-30
chore: use ethers6 js sdk in tools

idea404:
# What :computer: 
* Updates link in DevTools section to point to ethers6 JS SDK docs site instead of ethers5 version

# Why :hand:
* Was pointing to e5 version

# Evidence :camera:

image


MexicanAce:
Closing as no longer needed after docs shuffle

PR #776, opened 2023-11-03
QA - New wallet key line instructions for example section

obmatter:
# What :computer: 

# Why :hand:

# Evidence :camera:

# Notes :memo:


docs/dev/how-to/send-message-l2-l1.md
```diff
@@ -68,6 +68,8 @@ yarn add zksync-web3 ethers@5 typescript @types/node ts-node
 
 5. Create a `file.ts` file in the root directory with the next script:
 
+Note: Replace `<YOUR_PRIVATE_KEY>` with private key of the wallet to use.
```
MexicanAce, line 71: We should not recommend users type their Private Key in scripts. We should update this script to use the `.env` file to better ensure they don't accidentally share them or commit them into git

idea404 (changes requested):
Agree with @MexicanAce comment on adjusting text to exemplify using `.env` instead of hardcoding private information

docs/dev/how-to/send-message-l2-l1.md
obmatter: @MexicanAce fixed by adding instructions for .env file and also updating the scipt
image

obmatter: (For step 5 used text from another tutorials to remain docs look solid)

obmatter:
@MexicanAce @idea404 can you please review the PR with the already implemented fix for hardcoded wallet?
Thank you in advance!

idea404 (approved):

PR #777, opened 2023-11-03
fix: remove zks_getTokenPrice docs

Romsters:
# What :computer: 
- removed zks_getTokenPrice docs
- removed getTokenPrice SDKs methods docs

# Why :hand:
- zks_getTokenPrice will no longer be supported, it replaced with explorer APIs

# Evidence :camera:
Screenshot 2023-11-03 at 15 12 36


MexicanAce (approved):
LGTM

PR #802, opened 2023-11-30
fix: update tokens listing info

JackHamer09:
# What :computer: 
* Update "add token to bridge" info
* Add token listing FAQ

# Why :hand:
* Make it clear where Explorer and Bridge takes tokens info from

# Evidence :camera:
image
image


JackHamer09:
> how tokens will show on wallets

Well, I don't think there could be a single answer, it's up to each wallet's implementation

PR #804, opened 2023-11-30
Update providers.md

pekkadt04:
Hi, there is a typo; ZksAccounteBalances should be corrected to ZksAccountBalances in the output description.

# What :computer: 

# Why :hand:

# Evidence :camera:

# Notes :memo:


MexicanAce:
As this is a stale branch, I'm going to close, but I've opened a new PR to address this fix and will tag @pekkadt04 as the co-author

PR #807, opened 2023-12-01
fix: remove explorer reference from transaction statuses section

Romsters:
# What :computer: 
- remove explorer reference from transaction statuses section

# Why :hand:
- explorer doesn't show pending transactions


MexicanAce (approved):
LGTM

PR #820, opened 2023-12-11
fix: moved box-shadow to pseudo element

thomasvaeth:
# What :computer: 
* Changed the hover on `.card` from transitioning the box-shadow property to transitioning the opacity
* There were 2 opacity values on `.card` on the hover state (1 and 0.7) that were overriding each other, so I removed the 0.7 value

# Why :hand:
* Box-shadow should not be transitioned https://tobiasahlin.com/blog/how-to-animate-box-shadow/

# Evidence :camera:
Screenshot 2023-12-10 at 10 51 04 PM

# Notes :memo:
* VSCode automatically removes extra whitespace, so it looks like more changes than what it really is


MexicanAce:
Closing as this is a stale branch and docs shuffle made this change moot

PR #826, opened 2023-12-13
Spend Limit tutorial: new solidity version а or hardhat and env var for wallet private key

obmatter:
# What :computer: 
- solidity version required 0.8.20, but in the tutorial hardhat file we have 0.8.17 mentioned 
- in the deployFactoryAccount.ts file there’s a name for private key variable DEPLOYER_PRIVATE_KEY  but in the .env file from the downloaded template we have WALLET_PRIVATE_KEY= 
- 
# Why :hand:

# Evidence :camera:

# Notes :memo:


AlbionaHoti (approved):
LGTM

PR #845, opened 2024-01-11
Update bridging-asset.md

StanislavBreadless:
# What :computer: 

Adding a disclaimer that the tokens bridged by the default bridge do not support custom token functionality

# Why :hand:

To avoid accidental mistakes

# Evidence :camera:

image

# Notes :memo:


AlbionaHoti (approved):
LGTM

PR #855, opened 2024-01-16
fix: missing redirects 

AlbionaHoti:
# What :computer: 
* Fix missing redirects, example: `/dev/developer-guides/transactions/fee-model.html`
* Delete the folder of /docs/dev/how-to...

# Why :hand:
* There's a redirect for:  `/docs/dev/developer-guides/transactions/fee-model.html`
* No need for /docs/dev 

# Evidence :camera:

# Notes :memo:


MexicanAce (approved):
LGTM (can fix in a follow-up PR if necessary)

PR #857, opened 2024-01-16
feat(docs): add testnet faucets card to build

MexicanAce:
# What :computer: 
* Add `Testnet Faucets` card to `/build`

# Why :hand:
* Make it easier/faster to access network faucets

# Evidence :camera:
image


uF4No (approved):
🫡

PR #862, opened 2024-01-17
Adds zks_getProof to API docs

uF4No:
# What :computer: 
* Added `zks_getProof` info to API

# Why :hand:
* Method is currently undocumented
 
# Evidence :camera:
* See API page in preview

# Notes :memo:
Unlike other API methods, this one has a lot of info. The "Output" section is slightly different as it includes a description of the returned fields, which I think is required in this case as this method is quite complex.


MexicanAce (approved):
LGTM

PR #877, opened 2024-01-22
fix(python): fix python examples

petarTxFusion:
# What :computer: 
* Fix python examples links
* Update python examples


MexicanAce (approved):
LGTM

PR #884, opened 2024-01-24
fix: several typos

BoboTiG:
# What :computer: 
* Fix typos in markdown files
* Remove typos from cspell file

# Why :hand:
* For better user experience
* To ensure they don't continue in the future


MexicanAce (approved):
LGTM

PR #887, opened 2024-01-26
fix: updates ergs to gas

uF4No:
# What :computer: 
* Fixes a remaining `ergs` mention in the daily spend limit tutorial

# Why :hand:
* We changed from ergs to gas months ago

# Evidence :camera:
In preview

# Notes :memo:


MexicanAce (approved):
LGTM

PR #888, opened 2024-01-26
fix: build preview for upstream pull requests

MexicanAce:
# What :computer: 
* Allow users with `Write` access to manually trigger a deploy preview for upstream forks

# Why :hand:
* So that we can have higher confidence in changes coming from forks

# Notes :memo:
* Secrets should not be shared with upstream PRs and we require secrets for firebase deploys


AlbionaHoti (approved):
lgtm (I think)

PR #895, opened 2024-01-31
fix(python-sdk): fix typos and update package name

danijelTxFusion:
# What :computer: 
* Fix typos and package name.

# Why :hand:
* The name of the package is `zksync2` instead of `zksync2-python` which is the of the repository.


MexicanAce (approved):
LGTM

PR #898, opened 2024-02-02
docs: adds the SC memoization flag to standard JSON input

hedgar2017:
# What :computer: 

Adds the SC memoization flag to standard JSON input fields.

# Why :hand:

This flag can be used by various system projects to disable the memoization.

According to the Protocol team, sometimes it's not feasible to optimize out calls to SC, precompiles, etc.


MexicanAce (approved):
LGTM

PR #904, opened 2024-02-07
fix: fixed yarn command

amelnytskyi:
# What :computer: 
* Updated `yarn serve` command to `yarn dev` since `yarn serve` causes an error

# Why :hand:

# Evidence :camera:
image

# Notes :memo:


MexicanAce (approved):
LGTM, thank you!

PR #910, opened 2024-02-12
Note on paymaster transaction estimation

StanislavBreadless:
# What :computer: 

* Added a section on how to properly estimate gas when using paymaster in a transaction

# Why :hand:

* After the recent upgrade, the compute-related operations have become cheaper, while pubdata-related have become more expensive. We need to include a section on how to avoid missing additional pubdata-related costs when estimating transactions with a paymaster

# Evidence :camera:

# Notes :memo:


docs/build/developer-reference/account-abstraction.md
```diff
@@ -316,6 +316,42 @@ The paymaster supports only the [approval based](#approval-based-paymaster-flow)
 
 An example of how to use testnet paymaster can be seen in the [frontend quickstart](../tutorials/dapp-development/frontend-quickstart-paymaster.md#pay-fees-with-erc20-tokens) tutorial.
 
+### Estimating gas when interacting with a paymaster
+
+While the paymaster pays the fees for the user, it also consumes additional gas compared to a normal transaction, which includes: 
+
+1. Computation inside the paymaster's `validateAndPayForPaymasterTransaction` as well as `postTransaction`.
+2. Assigning and spending the allowance for the ERC20 tokens by the user.
+3. Paymaster sending funds to the bootloader.
```
uF4No, line 325:
```suggestion
2. Paymaster sending funds to the bootloader.
3. Assigning and spending the allowance for the ERC20 tokens by the user. (optional)
```

docs/build/developer-reference/account-abstraction.md
```diff
@@ -316,6 +316,42 @@ The paymaster supports only the [approval based](#approval-based-paymaster-flow)
...
+
+While the (1) is usually quite negligible (it, of course, depends on the implementation of each individual paymaster), (3) is roughly similar to what the user would pay in a transaction if it was to be spent by the user on its own, the (2) is unique for the paymaster operations, i.e. it is effectively an additional slot that is affected. Especially if the user grants allowance for the first time, since it would require 32 bytes to publish the storage key identifier, which may amount to roughly 400k gas under 50 gwei L1 gas price. Note, that even though the usual flow is "grant `X` allowance to the paymaster + paymaster spends all allowance", which means that the slot is zeroed out at the end of the execution, the cost for write is pre-charged during execution. In other words, we firstly charge for pubdata required for updating the allowance and if the slot has been zeroed out by the end of the transaction, the user will be refunded, but only at the end of the transaction.
```
uF4No, line 327: would need to adapt this paragraph if we change the order above

docs/build/developer-reference/account-abstraction.md
```diff
@@ -316,6 +316,42 @@ The paymaster supports only the [approval based](#approval-based-paymaster-flow)
...
+
+On the front-end side you should provide the corresponding `paymasterInput` to ensure that the paymaster is used during estimation. The corresponding snippet from the [Custom paymaster tutorial](../tutorials/smart-contract-development/paymasters/custom-paymaster-tutorial.md) shows how to do it:
+
+```ts
+  const gasLimit = await erc20.estimateGas.mint(wallet.address, 5, {
+    customData: {
+      gasPerPubdata: utils.DEFAULT_GAS_PER_PUBDATA_LIMIT,
+      paymasterParams: paymasterParams,
+    },
+  });
```
uF4No, line 339:
```suggestion
  const gasLimit = await greeterContract.estimateGas.setGreeting('A new message', {
    customData: {
      gasPerPubdata: utils.DEFAULT_GAS_PER_PUBDATA_LIMIT,
      paymasterParams: paymasterParams,
    },
  });
```
I think this example can be confusing as we talk about ERC20 approval before. I'd rather use another contract interaction to showcase estimation with paymaster params.

StanislavBreadless: Hmm, but the tutorial has exactly this example. I think it is simpler when the snippet is given within the context of a tutorial instead of just out of the blue

uF4No: that's ok. Let's discard this then

PR #919, opened 2024-02-15
Minor fix for missing quotation mark

kadirchan:
# What :computer: 
* The quotation marks were missing, I fix it

# Why :hand:
* It looked bad and was harder to copy

# Evidence :camera:
image


MexicanAce:
Thank you for the PR, looks like this was already merged upstream. Closing for now.

PR #920, opened 2024-02-15
fix: updated cli create project command

amelnytskyi:
# What :computer: 
* Updated cli create project command to `npx zksync-cli create [options]`
* Removed obsolete @latest tag after `zksync-cli`

# Why :hand:

# Evidence :camera:

# Notes :memo:


MexicanAce (approved):
LGTM, thank you!

PR #934, opened 2024-02-19
fix: fixing export statement being in 1 line

zk-Lumi:
# What :computer: 
* Fixed export statement for path being both on 1 line

# Why :hand:
* If not people might run command wrong


MexicanAce (approved):
LGTM

PR #939, opened 2024-02-20
fix: Empty contributors list on some pages

kelreel:
# What :computer: 
* Fixed a bug with formatting the list of contributors

# Why :hand:
* Not all pages contain a list of contributors. This led to an error (see console).

# Evidence :camera:
image
image


kelreel:
@dutterbutter locally, there is still such an error with env variablesб in the future, I think it's worth disabling SDK analytics in dev mode (I can do it later)

image

MexicanAce (approved):
LGTM, thank you!!

PR #940, opened 2024-02-21
Update bridges.md

MesProtocol:
Adding MES Protocol to the bridge section. We are offering a secure way to bridge assets across zkSync with other chains in an instant, low-cost and easy-to-use manner.


MexicanAce (approved):
LGTM, however I would recommend updating the logo used for `zkSync` to the latest version on your website.

PR #948, opened 2024-02-22
improved code formatting and timestamp calculation for time based paymasters

bug-author:
# What :computer: 
* code block formatting
* timestamp calculation

Both changes are for the [time based paymaster](https://docs.zksync.io/build/tutorials/smart-contract-development/paymasters/timebased.html#step-2-%E2%80%94-environment-setup) example page.

# Why :hand:
* code formatting for `endTime` line was a bit off
* the code does convey the intent for timestamp calculation, it was just not a valid/ correct calculation in terms of Solidity, I just tried to make it more clear

# Evidence :camera:
Before changes
```solidity
uint256 startTime = (block.timestamp / 86400) * 86400 + 15 hours;
 uint256 endTime = startTime + 20 minutes;

 require(
   block.timestamp >= startTime && block.timestamp <= endTime,
   "Transactions can only be processed between 14:35 - 14:55"
 );
```
image

After changes
```solidity
uint256 startTime = (block.timestamp / 86400) * 86400 + (15 * 3600); // Adding 15 hours (15 * 3600 seconds)
uint256 endTime = startTime + (20 * 60); // Adding 20 minutes (20 * 60 seconds)

 require(
   block.timestamp >= startTime && block.timestamp <= endTime,
   "Transactions can only be processed between 14:35 - 14:55"
 );
```
image


MexicanAce (approved):
LGTM, thank you for this

PR #950, opened 2024-02-22
fix: Update cross-chain tutorial to make use of latest deps and ethers v6 

dutterbutter:
# What :computer: 
* Updates to reflect latest dep changes from HH 
* Updates to make use of ethers v6
* Fixes CLI commands

# Why :hand:
* Should always use latest and greatest
* Issue with CLI command 

# Evidence :camera:

# Notes :memo:


vasyl-ivanchuk (approved):

vasyl-ivanchuk:
LGTM!

Issue #961, opened 2024-03-02
Error occurred during the deployment of "deploy-paymaster.ts" file.


yp945:
When I followed the tutorial document to deploy the paymaster contract, I found that there were errors in the content of the deploy-paymaster.ts file and some syntax related to `ethers` was problematic.

yp945:
I will try to fix it and create a pull request.

yp945:
> When I followed the tutorial document to deploy the paymaster contract, I found that there were errors in the content of the deploy-paymaster.ts file and some syntax related to `ethers` was problematic.

![image](https://github.com/matter-labs/zksync-web-era-docs/assets/15030378/6da5c1e7-1afe-42e2-92a6-d40db7711ee5)

PR #914, opened 2024-02-14
Fix tutorials to support zksync-ethers v6

uF4No:
# What :computer: 
* Fixing all tutorial scripts to adapt to the breaking changes introduced in `ethers` / `zksync-ethers` v6

- NFT Gated was already updated

# Why :hand:
* Most tutorial scripts are broken due to breaking changes in the new version of the libraries.

# Evidence :camera:
Screenshot 2024-02-14 at 16 02 42
Screenshot 2024-02-14 at 16 03 12

Screenshot 2024-02-14 at 16 04 21

# Notes :memo:
* Any notes/thoughts that the reviewers should know prior to reviewing the code?

PR #962, opened 2024-03-02
Fix paymaster deploy docs #961

yp945:
# What :computer: 
* First thing updated with this PR : Fix the error in obtaining the contract address.

# Why :hand:
* Reason why first thing was added to PR: The way to obtain the address of a deployed contract is problematic.

# Evidence :camera:
image

# Notes :memo:
Is it possible that other sample codes may also have issues with obtaining the contract address?


uF4No:
This will be included in #914

Issue #963, opened 2024-03-02
The document estimateGas method is used incorrectly.


yp945:
1. The gas estimation function in the `use-paymaster.ts` file contains errors.
2. bigInt multiplication error

![image](https://github.com/matter-labs/zksync-web-era-docs/assets/15030378/d93d2666-3e89-420a-bb01-43d60cdb3ef3)

sxf123333:
6

PR #964, opened 2024-03-02
fix: paymaster use docs #963

yp945:
# What :computer: 
* First thing updated with this PR : estimateGas function use error.
* Second thing updated with this PR: The calculation method of bigInt is incorrect.

# Why :hand:
* Reason why first thing was added to PR: Fixed the incorrect use of estimateGas.
* Reason why second thing was added to PR: Fixed the correct usage of bigInt multiplication.

# Evidence :camera:
image

# Notes :memo:
Is it because of different zksync-cli versions or ethers versions?


uF4No:
Thanks for the PR but this will be included in #914

PR #967, opened 2024-03-06
fix: change readme from yarn add to just yarn

zk-Lumi:
# What :computer: 
* Updated `yarn add` to `yarn`

# Why :hand:
* More clear this way


MexicanAce (approved):
LGTM

PR #968, opened 2024-03-06
Update blocks.md

savasadar:
Change block time interval value from 2 to 1.  L2 block is created every 1 seconds

# What 💻 
* blocks.md changed

# Why :hand:
* L2 block time is wrong

# Evidence :camera:


MexicanAce:
Confirmed the mainnet configuration is set to 1 second: `miniblock_commit_deadline_ms: 1000`

MexicanAce (approved):
LGTM

PR #981, opened 2024-03-20
fix: typos

miles-six:
I fix some typos and want to add workflow for codespell check to eliminate typos in the future.

I have added a relatively simple filtering file to exclude certain words. 

If you find it meaningful, you can modify the filtering conditions according to your needs.

The code quality is exceptionally high, so sincerely hope my PR can help make this repository more standardized.
# Evidence :camera:

# Notes :memo:


MexicanAce:
Hey @miles-six , thank you for the PR! We already have `cspell` in our linting workflow, but it appears to include a bad dictionary that's letting some misspellings through. Most have been caught while we investigate which dictionary is the problem, so for now I'm going to close the PR.

Apologies if this was negative work, but these type of investigations/feedback is very useful. And apologies for taking longer than expected to respond.

PR #983, opened 2024-03-21
feat: added a new sepolia faucet link

zm-moon:
- Added a new faucet link to the document
- Updated the document to include a total of 5 faucet links

This new faucet link is known for its reliability and ease of use compared to others.


MexicanAce:
As we already have the link to directly receive zkSync Sepolia ETH, there's no need for a Sepolia ETH faucet, as that would add extra steps to getting funds on the testnet

PR #986, opened 2024-03-25
docs(hardhat-other-plugins): add hardhat-preprocessor plugin

mpopovac-txfusion:
# What :computer: 
* Add hardhat-preprocessor plugin to the list of other plugins

# Why :hand:
* Introduce hardhat-preprocessor plugin with warning for contract verification

# Evidence :camera:
image


dutterbutter (dismissed):
LGTM just the one comment.

MexicanAce (approved):

PR #988, opened 2024-03-25
feat: add clave to wallets.md

zk-Lumi:
# What :computer: 
* Adding Calve to Wallets doc

# Why :hand:
* It is a first class UX and should be included
* Note: it is still in beta, but I think would be good to add here already


MexicanAce:
As there's still a wait-list to use their product, we should not direct users until it's available to everyone. Closing for now, but will keep an eye on when it's available for mass adoption

PR #1026, opened 2024-04-16
chore: updates foundry references in etn docs

dutterbutter:
# What :computer: 
* Removes outdated foundry related docs in etn docs

# Why :hand:
* The commands mentioned no longer are supported in favour of cast / forge

# Evidence :camera:

# Notes :memo:


MexicanAce (approved):
LGTM

PR #1031, opened 2024-04-25
fix: added warning about using private keys in plaintext

PatrickAlphaC:
# What :computer: 
Added a warning about using unencrypted private keys. 

# Why :hand:
New developers accidentally exposing private keys to GitHub is one of the quickest ways to destroy a new developer's motivation. Most people don't encrypt their keys and essentially just get lucky. We often don't hear about it because new developers are often embarrassed, and this goes underreported because the advice is, "Oh, you exposed your private key? Skill issue TBH." And then they leave Web3. A large part of this stems from the fact that most tutorials _encourage_ people to place their private keys in plaintext without any warning of the downsides. 

# Evidence :camera:

This poll shows that _most_ people are good, but it should be _all_. 

https://twitter.com/PatrickAlphaC/status/1783561089105465654

Out of the few that lose money, often lose a lot. 

https://twitter.com/PatrickAlphaC/status/1783562471602618712

We need to fix our tutorials so this is not the case. 

# Notes :memo:
* Related to https://github.com/matter-labs/zksync-docs/issues/187

### DIFF
--- a/docs/build/tutorials/smart-contract-development/account-abstraction/custom-aa-tutorial.md
+++ b/docs/build/tutorials/smart-contract-development/account-abstraction/custom-aa-tutorial.md
@@ -842,6 +842,12 @@ Make sure you deposit funds on zkSync Era using [one of the available bridges](h
 
 1. In the `deploy` folder, create the file `deploy-factory.ts` and copy/paste the following code, replacing `<WALLET_PRIVATE_KET>` with your private key.
 
+::: warning
+Remember to use a private key that only has testnet funds if you are going to paste your private key in plaintext. It is recommended to [never have a private key associated with real funds in plaintext.](https://github.com/Cyfrin/foundry-full-course-f23/discussions/5).
+
+If you wish to use hardhat with real funds, it's recommended to encrypt your key so it's not accidentally exposed.
+:::
+
 ```ts
 import { utils, Wallet } from "zksync-ethers";
 import * as ethers from "ethers";
--- a/docs/build/tutorials/smart-contract-development/account-abstraction/custom-aa-tutorial.md
+++ b/docs/build/tutorials/smart-contract-development/account-abstraction/custom-aa-tutorial.md
@@ -843,9 +843,9 @@ Make sure you deposit funds on zkSync Era using [one of the available bridges](h
 1. In the `deploy` folder, create the file `deploy-factory.ts` and copy/paste the following code, replacing `<WALLET_PRIVATE_KET>` with your private key.
 
 ::: warning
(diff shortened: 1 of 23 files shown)


docs/build/tutorials/smart-contract-development/account-abstraction/custom-aa-tutorial.md
```diff
@@ -842,6 +842,12 @@ Make sure you deposit funds on zkSync Era using [one of the available bridges](h
 
 1. In the `deploy` folder, create the file `deploy-factory.ts` and copy/paste the following code, replacing `<WALLET_PRIVATE_KET>` with your private key.
 
+::: warning
+Remember to use a private key that only has testnet funds if you are going to paste your private key in plaintext. It is recommended to [never have a private key associated with real funds in plaintext.](https://github.com/Cyfrin/foundry-full-course-f23/discussions/5).
+
+If you wish to use hardhat with real funds, it's recommended to encrypt your key so it's not accidentally exposed.
```
MexicanAce, line 848: Do you have a link to an example we can link to for encrypting your Private Key for use with something like Hardhat?

Also, as a general note, the upcoming new docs will lean more on tools like Foundry's `cast wallet import`to manage keys in a private encrypted keystore automatically for devs.

docs/build/tutorials/smart-contract-development/account-abstraction/custom-aa-tutorial.md
```diff
@@ -842,6 +842,12 @@ Make sure you deposit funds on zkSync Era using [one of the available bridges](h
 
 1. In the `deploy` folder, create the file `deploy-factory.ts` and copy/paste the following code, replacing `<WALLET_PRIVATE_KET>` with your private key.
 
+::: warning
```
MexicanAce, line 845: Let's please update all instances where this warning should be added. I believe the other tutorials also need this warning.

docs/build/tutorials/smart-contract-development/account-abstraction/custom-aa-tutorial.md
PatrickAlphaC: For Hardhat, you have to roll your own encryption script. They recommend using their variables setting, but I don't like it because the keys are still in plaintext and just moved elsewhere. 

[I have used this example in the past to encrypt and decrypt your own keys.](https://github.com/PatrickAlphaC/ethers-simple-storage-fcc/blob/main/encryptKey.js) The project has an example of encrypting and decrypting the keys. 

`cast wallet import` is the best way to do key management!! That's great! 

I'll add this example to the warning.

docs/build/tutorials/smart-contract-development/account-abstraction/custom-aa-tutorial.md
PatrickAlphaC: Done!

itsacoyote:
Moving Issue to zksync-docs repo and closing this PR. Will create a new PR for the changes in that repo.

PR #1039, opened 2024-05-01
Oracles

PatrickAlphaC:
# What :computer: 
* Added chainlink oracles to the oracles section

# Why :hand:
* Chainlink support price feeds on zksync


uF4No (approved):
🤘

PR #1058, opened 2024-05-07
Update accounts-l1-l2.md

githubdoramon:
# What :computer: 
* Fixing link to correct method
* 
# Why :hand:
* Cause it was wrong


uF4No:
can you also fix it in https://docs.zksync.io/build/sdks/js/accounts-l1-l2.html ??

uF4No (approved):
LGTM

